PR #7, opened 2019-01-31
Massive refactor of helix.vcl

drwilco:
This started out as a refactor to fix the handling of redirects from runtime for static files that are larger than the limit. It then quickly evolved to reworking how restarts are done, which is documented in the large comment at the top.

Other things were fixed along the way, mostly related, so not split out into separate changes:
 * Deduplicating code in `vcl_miss` and `vcl_pass`
 * Deduplicating `X-URL` and `X-Orig-URL`
 * Rename `X-Static` to `X-Request-Type`
 * Improve handling of proxy strains
 * Use less resources (headers/variables)
 * Be consistent about where to set the backend
 * Split `Image` out into a separate request type
 * Leak less data in headers to origins

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -70,14 +70,14 @@ sub hlx_check_from_edge {
 /**
  * Set the `X-From-Edge` header that the above sub checks.
  *
- * Should be called from the top of `vcl_recv`/`vcl_miss`
+ * Should be called from the top of `vcl_miss`/`vcl_pass`
  */
 sub hlx_set_from_edge {
   # If it exists, it's legit, leave it alone
   if (!bereq.http.X-From-Edge) {
     declare local var.data STRING;
     set var.data = strftime({"%s"}, now) + "," + server.datacenter;
-    set bereq.http.Secure-From-Edge =
+    set bereq.http.X-From-Edge =
       var.data + "," + digest.hmac_sha256(
         # This must match whatever key is being used 
         req.service_id + table.lookup(secrets, "OPENWHISK_AUTH"), var.data);
@@ -663,7 +663,8 @@ sub vcl_recv {
     # So we don't have to add `RESTART` to `X-Trace` whenever we use `restart`
     set req.http.X-Trace = req.http.X-Trace + "; RESTART; vcl_recv";
   }
-    
+
+  call hlx_set_from_edge;
   call hlx_recv_init;
 
 #FASTLY recv
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -38,9 +38,11 @@
 
 
 /**
- * Check and set the `X-From-Edge` header, which allows us to determine whether
+ * Check the `X-From-Edge` header, which allows us to determine whether
  * a request is coming from a Fastly edge POP. Use this instead of `Fastly-FF`
  * since that can be spoofed, or come from another Fastly service.
+ *
+ * This should be called from the very top of `vcl_recv`
  */
 sub hlx_check_from_edge {
   # Verify correct format
@@ -65,6 +67,11 @@ sub hlx_check_from_edge {
   unset req.http.X-From-Edge;
 }
 
+/**
+ * Set the `X-From-Edge` header that the above sub checks.
+ *
+ * Should be called from the top of `vcl_recv`/`vcl_miss`
+ */
 sub hlx_set_from_edge {
   # If it exists, it's legit, leave it alone
   if (!bereq.http.X-From-Edge) {
@@ -148,7 +155,7 @@ sub hlx_strain {
   }
 
   # Sanitize user input. `urlencode` leaves alphanumeric and `-._~`
-  set req.http.X-Strain = regsuball(urlencode(req.http.Foo), {"%.."}, "_");
+  set req.http.X-Strain = regsuball(urlencode(req.http.X-Strain), {"%.."}, "_");
 
   # do not override strain if set in header
   if (!req.http.X-Strain) {
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -10,6 +10,130 @@
 # governing permissions and limitations under the License.
 #
 
+
+# About URLs and Host headers:
+# This VCL will use the following conventions:
+# `req.url` - The canonical URL, will be used to determine the hash of the object
+# `req.http.X-Orig-URL` - The URL as received from the client
+# `req.http.X-Backend-URL` - The URL as should be sent to the origin (backend)
+#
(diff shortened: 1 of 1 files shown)


layouts/fastly/helix.vcl
```diff
@@ -22,23 +146,29 @@ sub hlx_strain {
   if (req.http.Cookie:X-Strain) {
     set req.http.X-Strain = req.http.Cookie:X-Strain;
   }
-  # we don't need cookies for anything else
-  unset req.http.Cookie;
+
+  # Sanitize user input. `urlencode` leaves alphanumeric and `-._~`
+  set req.http.X-Strain = regsuball(urlencode(req.http.Foo), {"%.."}, "_");
```
ejthurgo: `req.http.foo` = `req.http.X-Strain`?

layouts/fastly/helix.vcl
```diff
@@ -10,6 +10,130 @@
...
+        && time.is_after(time.add(now, 2s), var.time)) {
+      # Do nothing, the header is valid
+      return;
+    }
+  }
+  # Header isn't valid, chuck it
+  unset req.http.X-From-Edge;
+}
+
+sub hlx_set_from_edge {
```
ejthurgo: As someone looking at this VCL a fresh, personally I'd find it helpful if these subs had a bit better annotation and some context. it would help me better understand the intent if had a description and maybe even from where in the state machine it's intended to be called. Quite subjective though, so feel free to ignore.

layouts/fastly/helix.vcl
```diff
@@ -10,6 +10,137 @@
...
+    }
+  }
+  # Header isn't valid, chuck it
+  unset req.http.X-From-Edge;
+}
+
+/**
+ * Set the `X-From-Edge` header that the above sub checks.
+ *
+ * Should be called from the top of `vcl_recv`/`vcl_miss`
```
ejthurgo: Function accesses bereq. Should this be vcl_miss/vcl_pass?

layouts/fastly/helix.vcl
```diff
@@ -10,6 +10,137 @@
...
+ */
+sub hlx_set_from_edge {
+  # If it exists, it's legit, leave it alone
+  if (!bereq.http.X-From-Edge) {
+    declare local var.data STRING;
+    set var.data = strftime({"%s"}, now) + "," + server.datacenter;
+    set bereq.http.Secure-From-Edge =
```
ejthurgo: Should this be `X-From-Edge`

layouts/fastly/helix.vcl
```diff
@@ -10,6 +10,137 @@
...
+  # Header isn't valid, chuck it
+  unset req.http.X-From-Edge;
+}
+
+/**
+ * Set the `X-From-Edge` header that the above sub checks.
+ *
+ * Should be called from the top of `vcl_recv`/`vcl_miss`
+ */
+sub hlx_set_from_edge {
```
ejthurgo: In fact, I'm not sure this is ever called?

trieloff:
@drwilco thank you for the PR. I'll wait for @ejthurgo's points to be addressed before I merge.

layouts/fastly/helix.vcl
tripodsan: @drwilco can we delete this function?

layouts/fastly/helix.vcl
tripodsan: @drwilco I think @ejthurgo is correct. should we rename this?

trieloff:
@tripodsan I spoke with @ejthurgo last week and he said that he wasn't entirely done with his review.

layouts/fastly/helix.vcl
```diff
@@ -10,6 +10,137 @@
...
+# `req.http.X-Backend-URL` - The URL as should be sent to the origin (backend)
+#
+# Having `req.url` be canonical, vs `X-Backend-URL` being what's sent to origin 
+# allows us to do restarts to find content on other origins, and use origin
+# specific URLs, yet have the object stored under a single hash, meaning that 
+# a cache lookup will find the proper object on the first attempt once an 
+# object is in the cache.
+#
+# So `req.url` is what should be left after filtering query string parameters,
+# doing normalizations, etc. Then `req.http.X-Backend-URL` is a transformation of
```
ejthurgo: Is it possible that backend transformation could result in any normalization. and therefore cache efficiency?

layouts/fastly/helix.vcl
```diff
@@ -636,77 +841,43 @@ sub vcl_fetch {
...
   }
   return(deliver);
 }
 
-sub vcl_miss {
-#FASTLY miss
-  set req.http.X-Trace = req.http.X-Trace + "; vcl_miss";
-  unset bereq.http.X-Orig-Url;
+/**
+ * Do all BackEnd REQuest changes. One sub to be called from both vcl_miss and
```
ejthurgo: `BackEnd REQuest`
`backend request`

layouts/fastly/helix.vcl
```diff
@@ -636,77 +841,43 @@ sub vcl_fetch {
...
   if (req.backend.is_shield) {
     set bereq.url = req.http.X-Orig-Url;
-  }
-
-  # set backend host
-  if (req.backend == F_AdobeRuntime) {
-    set bereq.http.host = "adobeioruntime.net";
-  } elsif (req.backend == F_GitHub) {
-    set bereq.http.host = "raw.githubusercontent.com";
+    set bereq.http.Host = req.http.X-Orig-Host;
```
ejthurgo: we stash the original host in `req.http.X-Orig-Host`, however, we never actually change `req.http.host`
Is this risk management against future changes to host in vcl_recv, or just an oversight/relic from an older version.

layouts/fastly/helix.vcl
ejthurgo: It's not superfluous, we just need to reference it.
I believe it should be called here: https://github.com/adobe/helix-publish/blob/c9c4a0f30f9d3faa37718024a244138e5b29b8e4/layouts/fastly/helix.vcl#L870

layouts/fastly/helix.vcl
drwilco: It shouldn't. All normalization should be done on `req.url` and `req.http.Host`.

layouts/fastly/helix.vcl
drwilco: I did that capitalization to explain where `bereq` comes from :)

layouts/fastly/helix.vcl
drwilco: Risk management vs future changes

layouts/fastly/helix.vcl
drwilco: Correct. Adding.

PR #27, opened 2019-03-07
Draft for redirect management

trieloff:
Another step for implementing https://github.com/adobe/helix-cli/issues/587

Basic idea:

- `helix-publish` will create a number of conditions and headers that set the `req.http.X-Location` header when the request URL matches a pattern specified in the `redirects[].from` property of `helix-config.yaml` (see https://github.com/adobe/helix-shared/blob/master/docs/redirectrule.schema.md)
- `vcl_recv` will check if `X-Location` is set and throw an error 301
- the 301 error handler will create a synthetic response, with the status code 301 and copy `req.http.X-Location` to `resp.http.Location`, forcing the redirect

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -674,6 +674,12 @@ sub vcl_recv {
 
 #FASTLY recv
 
+  # X-Location gets set by Helix's redirect logic inside the FASTLY recv block
+  # that will get injected above.
+  if (req.http.X-Location) {
+    error 301 "Redirect"
+  }
+
   # TODO: Do we even want to set a regular origin as default? Possibly set one
   # that acts as a canary. If requests reach it, the VCL isn't setting a backend
   # somewhere.
@@ -1000,6 +1006,13 @@ sub vcl_deliver {
 sub vcl_error {
   set req.http.X-Trace = req.http.X-Trace + "; vcl_error";
 #FASTLY error
+
+  if (obj.status == 301) {
+    set obj.http.Content-Type = "text/html";
+    set obj.http.Location = req.http.X-Location;
+    synthetic "Moved Permanently";
+    return(deliver);
+  }
   call hlx_error_errors;
 }
 
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -677,7 +677,7 @@ sub vcl_recv {
   # X-Location gets set by Helix's redirect logic inside the FASTLY recv block
   # that will get injected above.
   if (req.http.X-Location) {
-    error 301 "Redirect"
+    error 301 "Redirect";
   }
 
   # TODO: Do we even want to set a regular origin as default? Possibly set one
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -1010,7 +1010,7 @@ sub vcl_error {
   if (obj.status == 301 && req.http.X-Location) {
     set obj.http.Content-Type = "text/html";
     set obj.http.Location = req.http.X-Location;
-    synthetic "Moved Permanently";
+    synthetic "Moved permanently <a href='" + req.http.X-Location+ "'>" + req.http.X-Location + "</a>";
     return(deliver);
   }
   call hlx_error_errors;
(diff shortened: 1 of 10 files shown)


layouts/fastly/helix.vcl
```diff
@@ -674,6 +674,12 @@ sub vcl_recv {
 
 #FASTLY recv
 
+  # X-Location gets set by Helix's redirect logic inside the FASTLY recv block
+  # that will get injected above.
+  if (req.http.X-Location) {
+    error 301 "Redirect"
```
ejthurgo: missing `;`

layouts/fastly/helix.vcl
```diff
@@ -1000,6 +1006,13 @@ sub vcl_deliver {
 sub vcl_error {
   set req.http.X-Trace = req.http.X-Trace + "; vcl_error";
 #FASTLY error
+
+  if (obj.status == 301) {
+    set obj.http.Content-Type = "text/html";
+    set obj.http.Location = req.http.X-Location;
+    synthetic "Moved Permanently";
```
ejthurgo: Nothing _really_ wrong with this because most clients will automatically redirect, but spec does say:
> Unless the request method was HEAD, the entity of the response SHOULD contain a short hypertext note with a hyperlink to the new URI

I'd be tempted to save some bytes and give an empty response, or do as per spec, or feel free to ignore this comment.

layouts/fastly/helix.vcl
```diff
@@ -1000,6 +1006,13 @@ sub vcl_deliver {
 sub vcl_error {
   set req.http.X-Trace = req.http.X-Trace + "; vcl_error";
 #FASTLY error
+
+  if (obj.status == 301) {
```
ejthurgo: Overriding all 301s with X-Location is perhaps overzealous.
Because this is fairly specific functionality, perhaps error with a unique code (9XX) so this code block only handles what we want and then rewrite the status code.

trieloff:
```suggestion
  if (obj.status == 301 && req.http.X-Location) {
```

maybe that's better, @ejthurgo

layouts/fastly/helix.vcl
```suggestion
    synthetic "Moved permanently <a href='" + req.http.X-Location+ "'>" + req.http.X-Location + "</a>";
```

Would this satisfy the spec?

layouts/fastly/helix.vcl
trieloff:
```suggestion
    error 301 "Redirect";
```

layouts/fastly/helix.vcl
trieloff:
```suggestion
    synthetic "Moved permanently <a href='" + req.http.X-Location+ "'>" + req.http.X-Location + "</a>";
```

Would this satisfy the spec?

trieloff:
Now includes some work for #30

test/vcl-utils.test.js
```diff
@@ -79,4 +79,16 @@ describe('Testing vcl-utils.js', () => {
...
+  it('#pattern2vcl', () => {
+    assert.equal(utils.pattern2vcl('/foo/bar'), '"/foo/bar"');
+    assert.equal(utils.pattern2vcl('/foo/$1.html'), '"/foo/" + re.group.1 + ".html"');
+    assert.equal(utils.pattern2vcl('/foo/$1/bar/$2.html'), '"/foo/" + re.group.1 + "/bar/" + re.group.2 + ".html"');
+    assert.equal(utils.pattern2vcl('/foo/$1/bar/$2.$3'), '"/foo/" + re.group.1 + "/bar/" + re.group.2 + "." + re.group.3 + ""');
+  });
+
+  it('#condition', () => {
+    assert.equal(utils.condition('/old/(.*)', 'default'), 'req.http.X-Strain == "default" && req.url ~ "/old/(.*)"');
+    assert.equal(utils.condition('https://(.*)\.adobe.io', 'staging'), 'req.http.X-Strain == "staging" && ("https://" + req.http.host + req.url) ~ "https://(.*).adobe.io"');
```
trieloff: This is assuming that `("https://" + req.http.host + req.url) ~ "https://(.*).adobe.io"` is valid VCL, i.e. that VCL regex can be used on calculated values.

/cc @ejthurgo

trieloff:
Closing in favor of #34

trieloff:
Closing in favor of #34

PR #29, opened 2019-03-08
Only append params if there are params

### LINKED ISSUE #23: payload.request.url comes with an ugly "?(null)" at the end on "publish"
`hlx publish` the demo site and output `payload.request.url` for a request to `/index.html`: you get `/index.html?(null)`
You would expect to get `/index.html`

Interestingly, requesting `/index.html?test=true` also outputs `/index.html?(null)` while you would at least expect ``/index.html?test=true`.

trieloff:
Fixes #23


trieloff:
This should probably wait for #34

Issue #59, opened 2019-04-18
Directory Index is never set


trieloff:
The property for `strain_index_files` is never set.

PR #60, opened 2019-04-18
fix(dictionaries): set missing directory index

trieloff:
See #59


trieloff:
@kptdobe you can test this by running `hlx publish --api-publish https://adobeioruntime.net/api/v1/web/helix/default/publish-test`

kptdobe:
Awesome, worked! See https://helix.adobelaunch.com/ which shows the README as expected!

PR #61, opened 2019-04-18
Enable Better Caching for Static Resources

trieloff:
provide two new resource types for static resources: .esi to replace a static URL with a cacheable static URL and .hlx_[0-9a-f]+ for resources that are cached forever in the browser

In order to increase the cache efficiency of static resources, we are implementing a multi-stage process which replaces the generic, short and uncacheable URLs of static resources like `/scripts/site.js` with long and version-specific URLs like `/scripts/site.js.hlx_f7c3bc1d808e04732adf679965ccc34ca7ae3441`. This is done by offering a virtual resource like `/scripts/site.js.esi` that can be included in an ESI tag like `<script src="<esi:include src="/scripts/site.js.esi" />">` from within HTML, JS or CSS (the replacement of `.js` and `.css` file endings iwth `.js.esi` and `.css.esi` will be handled in helix-pipeline and helix-static at a later point in time). This virtual resource will then simply replace the short URL `/scripts/site.js` with the long-cachable URL `/scripts/site.js.hlx_f7c3bc1d808e04732adf679965ccc34ca7ae3441` through the ESI resolution process. Any static resource requested with a long-cachable URL will be delivered with Cache-Control headers setting expiry 366 days into the future.

Partial implementation of https://github.com/adobe/helix-pipeline/issues/224#issuecomment-476690621
Requires https://github.com/adobe/helix-cli/pull/797 and https://github.com/adobe/helix-pipeline/pull/281

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -549,9 +549,10 @@ sub hlx_fetch_static {
 }
 
 sub hlx_deliver_static {
-  set req.http.X-Trace = req.http.X-Trace + "; hlx_deliver_static";
+  set resp.http.X-Static-Trace = req.http.X-Trace + "; hlx_deliver_static";
+  set resp.http.X-Static-Trace = set resp.http.X-Static-Trace + "[type=" + req.http.X-Request-Type + ", status=" + resp.status + "]";
   if (req.http.X-Request-Type == "Static-ESI" && resp.status == 200) {
-    set req.http.X-Trace = req.http.X-Trace + "(esi)";
+    set resp.http.X-Trace = resp.http.X-Trace + "(esi)";
     # Get the ETag response header and use it to construct a stable URL
     declare local var.ext STRING;
 
@@ -1071,9 +1072,9 @@ sub vcl_deliver {
   set req.http.X-Trace = req.http.X-Trace + "; vcl_deliver";
 #FASTLY deliver
 
-  call hlx_deliver_static;
-
   call hlx_headers_deliver;
+
+  call hlx_deliver_static;
 
   # only set the strain cookie for sticky strains
   # and only do it for the Adobe I/O Runtime backend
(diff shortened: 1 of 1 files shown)


trieloff:
You can see the entire thing in action here: https://www.primordialsoup.life/index.html

layouts/fastly/helix.vcl
```diff
@@ -453,14 +527,50 @@ sub hlx_type_redirect {
...
+  # check for hard-cached files like /foo.js.hlx_f7c3bc1d808e04732adf679965ccc34ca7ae3441
+  if (req.http.X-Orig-URL ~ "^(.*)(.hlx_([0-9a-f]){20,40}$)") {
+    set req.http.X-Trace = req.http.X-Trace + "(immutable)";
+
+    set var.ext = ".hlx_" + digest.hash_sha1(beresp.http.ETag);
+
+    if (re.group.2 == var.ext) {
+      set req.esi = true;
+      esi;
+      set beresp.http.X-ESI = "processed(" + bereq.http.Accept-Encoding + ", " + req.esi + "," + req.http.X-From-Edge +")";
```
```suggestion
      set beresp.http.X-ESI = "processed(" + bereq.http.Accept-Encoding + ", " + req.esi + "," + req.http.X-From-Edge +")";
      `set beresp.http.x-compress-hint = "on";` // so h2o compresses it on the way out
```

As recommended by @vvuksan

ejthurgo:
Appreciate this is just semantics: but why `.esi`? 
Besides the fact it’s using ESI to achieve it, I’m not sure I associate ESI with what the feature is actually doing. Why not something like `.static` or `.immuatable` or similar to signify that you want the "versionless" entity.

trieloff:
@ejthurgo I've been considering `.url` as the extension, as it contains an URL. Later on, I'd also create a variant of the resource that redirects to the immutable resource and would probably pick the extension `.302`.

PR #71, opened 2019-05-06
Proxy Strain Subdirectory Support

### LINKED ISSUE #64: Proxy strains does not work on URL path
**Description**
Proxy strain works for host but not on the full URL path.

**To Reproduce**
  - name: support-production
    origin: https://adobedocs.github.io/developer-support/
    url: https://developer.adobe.com/support

**Expected behavior**
fetching origin from https://adobedocs.github.io/developer-support/

**Screenshots**
fetches https://adobedocs.github.io/support
image

**Version:**
1.0.0

**Additional context**
Current proxy strain creates rule for host but doesn't take path into consideration.

  set req.http.host = "adobedocs.github.io";

it should be:
set req.http.host = "adobedocs.github.io";
set req.url = "/developer-support/";

trieloff:
Fixes #64 by rewriting `https://helixdomain.com/newpath` to `https://backenddomain.com/oldpath` using the `strain[].origin.path` property and the `strain.url.path` property.

Note: this is a breaking change that changes the behavior of all proxy strains. If you had a proxy strain for `https://helixdomain.com/foobar/` then requests to `https://helixdomain.com/foobar/index.html` would go go `https://backenddomain.com/foobar/index.html`, now they go to `https://backenddomain.com/index.html`. To restore the old behavior, add a `path` to each `origin` in a proxy strain.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,9 +15,9 @@
       }
     },
     "@adobe/helix-shared": {
-      "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/@adobe/helix-shared/-/helix-shared-1.0.2.tgz",
-      "integrity": "sha512-E3V+4FgCPR5mYW72eyKkAe/85gv3XYn56GhpOKJfdU7aZyoo3NJ93yZZta67gifoEs7uNIy5d5LIZMmBqQUKxA==",
+      "version": "1.1.0",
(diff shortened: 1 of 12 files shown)


trieloff:
@hirenoble you can try this by running 
```bash
$ hlx publish --api-publish https://adobeioruntime.net/api/v1/web/helix/default/publish-test
```

Issue #85, opened 2019-05-14
Add extensions.vcl and the type_pipeline extension points


kptdobe:
For adobe/helix-cli#812 and as discussed there, we need: 

- [x] add a `extensions.vcl` file 
- [ ] the file will contain 2 empty subroutines:
  - [x] hlx_type_pipeline_before 
  - [ ] hlx_type_pipeline_after
- [x] the `extensions.vcl` is included by `helix.vcl`
- [x] the 2 new subroutines are called at the beginning and the end of hlx_type_pipeline subroutine
- [x] the `extensions.vcl` file is upload to Fastly exactly like the other vcl files

The override capability is handled in https://github.com/adobe/helix-publish/issues/79

kptdobe:
While writing the code, I think `extensions.vcl` makes more sense than `custom.vcl`.

Issue #79, opened 2019-05-10
Support Custom VCL


trieloff:
See https://github.com/adobe/helix-cli/issues/812#issuecomment-491186800

The payload should accept a new parameter `vcl`, which is an object. For each property of that object that has a corresponding file in `/layouts/fastly/${name}.vcl`, create a VCL file using the Fastly API, which is overriding the default from `/layouts/fastly/*`

kptdobe:
To cover adobe/helix-cli#812, we only need to be able to override the newly introduced `custom.vcl` (see #85). We do not have a use yet for all the others, I would not offer the capability and wait for customer asking for it.

kptdobe:
Custom VCL support has been introduced via the vcl https://github.com/adobe/helix-publish/blob/master/layouts/fastly/extensions.vcl. 
The publish service accepts now a `vcl` parameter and for now only support "override" is this `extensions.vcl` that can be overriden by providing the value: `{ ... "vcl": { "extensions": "your_extensions_vcl_override"}...}`

To use in a project, check https://github.com/adobe/helix-cli/issues/812.

PR #86, opened 2019-05-14
feat(vcl): Add extensions.vcl

kptdobe:
PR for https://github.com/adobe/helix-publish/issues/85

### DIFF
--- a/layouts/fastly/extensions.vcl
+++ b/layouts/fastly/extensions.vcl
@@ -10,6 +10,10 @@
 # governing permissions and limitations under the License.
 #
 
-sub hlx_type_pipeline_before {}
+sub hlx_type_pipeline_before {
 
-sub hlx_type_pipeline_after {}
+}
+
+sub hlx_type_pipeline_after {
+    
+}
--- a/layouts/fastly/extensions.vcl
+++ b/layouts/fastly/extensions.vcl
@@ -1,0 +1,15 @@
+#
+# Copyright 2018 Adobe. All rights reserved.
+# This file is licensed to you under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License. You may obtain a copy
+# of the License at http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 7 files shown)


layouts/fastly/extensions.vcl
```diff
@@ -0,0 +1,15 @@
+#
+# Copyright 2018 Adobe. All rights reserved.
```
trieloff:
```suggestion
# Copyright 2019 Adobe. All rights reserved.
```

src/fastly/vcl.js
```diff
@@ -26,6 +26,12 @@ async function init(fastly, version) {
   return fastly.setMainVCL(version, 'helix.vcl');
 }
 
+async function extensions(fastly, version) {
```
trieloff: How do you pass the customizations?

kptdobe: This will be part of https://github.com/adobe/helix-publish/issues/79 (clear separation of code ;) )

kptdobe:
> Make sure to increase the number 6 to 7.

I assume you talk about a version number but where / when this increase should happen ?

Also, the tests are failing because of code coverage. But this seems to be inherited from `master`... why ? if code cover target is not reached and it is ok then we need to change the check to be green. Or... ?

trieloff:
The only test that is failing on the CI is this:

Integration Test Test publish function locally - Test publish function locally
`/home/circleci/project/test/integration.js`
```
AssertionError [ERR_ASSERTION]: { body: { status: 'published', completed: 7 }, statusCode: 200 } deepStrictEqual { body: { status: 'published', completed: 6 }, statusCode: 200 }
    at Context.condit (test/integration.js:108:12)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
```

If you check the expected value, your tests should pass on the CI and your code coverage should be ok, too. Code coverage is lower when running locally without Fastly credentials because the tests that require these credentials get skipped.

kptdobe:
Aggggrrr... I feature we cannot define subroutines in an external file! Need to find another way...

PR #87, opened 2019-05-14
feat(vcl): Support Custom VCL

kptdobe:
PR for #79

PR #91, opened 2019-05-15
Helix Pages support code

koraa:
Will be merged thursday around noon

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -210,9 +210,14 @@ sub hlx_block_recv {
 sub hlx_owner {
   set req.http.X-Trace = req.http.X-Trace + "; hlx_owner";
 
-  set req.http.X-Owner = table.lookup(strain_owners, req.http.X-Strain);
-  if (!req.http.X-Owner) {
-    set req.http.X-Owner = table.lookup(strain_owners, "default");
+  if (req.http.host ~ "project-helix.dev$") {
+    # For helix pages we calculate the repo to use based on the url
+    set req.http.X-Owner = regsub(regsub(req.http.host, "\..*$", ""), "^.*-", "");
+  } else {
+    set req.http.X-Owner = table.lookup(strain_owners, req.http.X-Strain);
+    if (!req.http.X-Owner) {
+      set req.http.X-Owner = table.lookup(strain_owners, "default");
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


trieloff:
Looks good, but I'd prefer this to simply introduce the right extension mechanisms so that it works with #87, i.e. have callbacks in `hlx_owner`, etc.

koraa:
I've looked at the extension points; right now they are not powerful enough; I suggest we use the new code for helix pages support as a basis to design the extension support around. This way, we can get rid of special cases one by one.

kptdobe:
-1
Add project specific code inside the product is not nice, especially now that we are about to have an extension point mechanism.

> I suggest we use the new code for helix pages support as a basis to design the extension support around.

That's exactly how I proceed. I am adding the only extension point (and all the machinery around to support it) I need to cover "my" current needs. Once we move forward, we can add more. Feel free to open an issue for the extension point(s) you would need.

koraa:
It's not really project specific as helix pages will replace the helix-cli way of deploying any helix pages; since helix pages is a very core aspect of helix, it should not be put into an extra repo IMO.
I'd be happy to utilize your extension functions @kptdobe, however they are not up to the tasks; I did have a look at them before writing this PR.

@kptdobe Could you please demonstrate how the functionality of helix pages could be achieved with your extension points? If it's not possible I would like to go ahead with this as helix pages is high priority.

rofe:
> helix pages will replace the helix-cli way of deploying any helix pages

When was that decision taken? Is it persisted somewhere? I thought that subject still needs to be discussed, hence the breakout session at the next hackathon...

kptdobe:
Let discuss about this PR next week at the hackathon.

kptdobe:
> helix pages will replace the helix-cli way of deploying any helix pages

@koraa it is not on your roadmap, @davidnuescheler removed this confusing sentence:
https://github.com/adobe/project-helix/commit/70e5ecfb18967084f8897b57837c887e03018b64#diff-334f2a839b7a966b08f306945f7a76fc

kptdobe:
> [extension functions] however they are not up to the tasks

In the current helix-pages repo, I created the required extensions to use the subdomain to compute the owner / repo / ref based on the pattern:
https://github.com/adobe/helix-pages/pull/12/files

You can inspire from that and if an extension point is missing, you can make a PR to add it. See:
https://github.com/adobe/helix-publish/pull/93

Issue #96, opened 2019-05-16
Refactor Helix Flow to give Content Repository Precedence for all Request


trieloff:
I've been discussing this with @davidnuescheler and since we have the new drawings in docs (https://github.com/adobe/helix-publish/tree/master/docs), I can outline the change I'd like to achieve.

Old:
 
<img width="1039" alt="Helix Flow@2x" src="https://user-images.githubusercontent.com/39613/57866648-f73f5700-77ff-11e9-8d13-87d11407aa0a.png">

New:

<img width="1303" alt="Helix Flow New@2x" src="https://user-images.githubusercontent.com/39613/57866647-f6a6c080-77ff-11e9-96db-595f7c5349fa.png">

What changes:

1. all normal requests get the `raw` request type first (this used to be called `image)
2. only when the extension is supported by the image optimizer, we will enable image optimization
3. otherwise the `raw` request type works like the old `image` type, it tries to fetch the content directly from the content repository, even when it is a `html`, `js`, or `css` file. 
  - this means you can create a static `index.html` and get it served @filmaj 
  - this means you can have custom CSS for Helix Pages @koraa and @kptdobe 
  - this means you can have a `404.html` @tripodsan @ramboz (but more about that later)
4. when this request (which is very cheap for us) fails, we fall back to the regular `pipeline` request type, i.e. try to call the runtime action.
  - note: the most frequent failure cases, i.e. no action for `css`, `js`, etc. is very cheap and fast for us, because no action gets called
5. when the `pipeline` request fails, we fall back to `static`, which can (as before yield content, or a `redirect`, or a `static_url`, each with a restart.
6. when that request fails, we start a new request type `error`, which changes the internal URL to `404.html` and tries to get it from the content repository

Examples of requests and resolution:

| Request | Content repo | Static repo  | Comment |
|:-------- |:--------------|:------------|:----------|
| /index.html     | index.md :white_check_mark: | | Default md processing |
| /index.html     | index.html :white_check_mark:; index.md     | | Static html file stronger than md processed file|
| /index.html     | index.html :white_check_mark: | index.html | |
| /index.html     | index.html :white_check_mark: | htdocs/index.html | |
| /index.html     | | htdocs/index.html :white_check_mark: | |
| /index.html     | | htdocs/index.md :| ❌ 404 - no pipeline processing from static |
| /index.html     | | index.html :| ❌ 404 - static action mapped to `htdocs` |
[...]

PR #103, opened 2019-05-29
Implement new Request Flow

trieloff:
See #96 

#### Update

We now have two competing implementations of #96:
- #103 (Fastly-based, sequential requests)
- #116 (OpenWhisk-based, parallel requests

You can try this version by running `hlx publish --api-publish https://adobeioruntime.net/api/v1/web/helix/helix-services/publish@ci688` (replace the `688` with newer CircleCI build numbers if needed)


trieloff:
- [ ] raw
- [ ] pipeline
- [ ] static
- [ ] url
- [ ] `404.html`
- [ ] 404 fallback
- [ ] image optimization

trieloff:
Depending on the result of the investigation in https://github.com/adobe/helix-pipeline/issues/365 this will require an additional change.

PR #116, opened 2019-06-24
Dispatch flow

trieloff:
This is yet another implementation of #96. It is based on, but conflicts with #103 to we have to pick either, nor neither one. 

Other than #103 which runs sequential requests in Fastly, this PR runs parallel requests from a wrapper action in OpenWhisk (https://github.com/adobe/helix-experimental-dispatch), which adds about 20ms overhead in the best case scenario (requested resource exists in the content repository), but performs much better as soon as the first fallback is tried.

This is still a draft, as there is some cleanup work to do:

- [x] `params` support https://github.com/adobe/helix-experimental-dispatch/issues/5
- [x] cleanup unused functions from VCL
- [x] fix VCL errors
- [x] testing

#### Update

We now have two competing implementations of #96:
- #103 (Fastly-based, sequential requests)
- #116 (OpenWhisk-based, parallel requests

You can try this version by running `hlx publish --api-publish https://adobeioruntime.net/api/v1/web/helix/helix-services/publish@ci731` (replace the `689` with newer CircleCI build numbers if needed)


trieloff:
Approved in this call: https://bluejeans.com/s/cAUYt/

Issue #38, opened 2018-10-18
Protect X-Debug header with a key


trieloff:
Agree. Storing it in a dictionary is just an extra step that we don't need to take.

I'd always print the debug key, but allow setting it with `--debug-key=hunter2`

_Originally posted by @trieloff in https://github.com/adobe/helix-cli/issues/264#issuecomment-430579963_

trieloff:
I think we could simply use the service config ID, which is well-known to the developer, but not otherwise shared. @davidnuescheler what do you think?

kptdobe:
I am going to take a look at this with @MarquiseRosier to explore with him this part of the stack.

PR #177, opened 2019-09-18
feat(pub): protect X-Debug with key

MarquiseRosier:
https://github.com/adobe/helix-publish/issues/38

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -112,7 +112,6 @@ sub hlx_recv_init {
   set req.http.X-Orig-URL = req.url;
   set req.http.X-Orig-Host = req.http.Host;
 
-  # TODO: Make X-Debug require a valid token
   if (!req.http.X-Debug) {
     # Make sure clients can't change request flow
     # X-Strain is currently settable by Cookie, not worth unsetting for now
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -801,11 +801,11 @@ sub hlx_check_debug_key {
   declare local var.level STRING;
 
   set var.debugSecret = table.lookup(secrets, "DEBUG_KEY");
-  set var.key = regsub(req.http.X-Debug, ":.*", "");
+  set var.key = regsub(req.http.X-Debug, ":.*$", "");
 
-  if(req.http.X-Debug ~ ":"){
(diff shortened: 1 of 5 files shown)


layouts/fastly/helix.vcl
```diff
@@ -793,6 +792,31 @@ sub hlx_type_proxy {
...
+/**
+  This checks X-Debug key
+*/
+sub hlx_check_debug_key {
+  declare local var.debugSecret STRING;
+  declare local var.key STRING;
+  declare local var.level STRING;
+
+  set var.debugSecret = table.lookup(secrets, "DEBUG_KEY");
+  set var.key = regsub(req.http.X-Debug, ":.*", "");
```
```suggestion
  set var.key = regsub(req.http.X-Debug, ":.*$", "");
```

just in case: match until the end of the string.

layouts/fastly/helix.vcl
```diff
@@ -793,6 +792,31 @@ sub hlx_type_proxy {
...
 
+/**
+  This checks X-Debug key
+*/
+sub hlx_check_debug_key {
+  declare local var.debugSecret STRING;
+  declare local var.key STRING;
+  declare local var.level STRING;
+
+  set var.debugSecret = table.lookup(secrets, "DEBUG_KEY");
```
trieloff: You could specify a default value here as a third parameter, for instance the service config ID.

layouts/fastly/helix.vcl
```diff
@@ -793,6 +792,31 @@ sub hlx_type_proxy {
...
+  set var.key = regsub(req.http.X-Debug, ":.*", "");
+
+  if(req.http.X-Debug ~ ":"){
+    set var.level = regsub(req.http.X-Debug, ".*:", "");
```
```suggestion
    set var.level = regsub(req.http.X-Debug, "^.*:", "");
```

same idea: match from the start of the string

layouts/fastly/helix.vcl
```diff
@@ -793,6 +792,31 @@ sub hlx_type_proxy {
...
+  }else {
```
```suggestion
  } else {
```

whitespace fix

layouts/fastly/helix.vcl
MarquiseRosier: Ahh for here; this is done in the CLI! the publish always either sends serviceid or some custom debug-key so it will be in the table

trieloff: Sounds good and is probably safer, too.

PR #264, opened 2019-12-20
feat(dispatch): update to dispatch@v3 with dynamic defaults

trieloff:
BREAKING CHANGE: This changes the default dispatch version from v2 to v3. v3 introduces dynamic defaults (fallback to `default.htnl` in the same folder if file can't be found), which is a breaking change

enables dynamic defaults, see https://github.com/adobe/helix-dispatch/issues/138

**Note:** as this is a breaking change, it will require a new helix-cli version, too.


kptdobe:
Let's cross the fingers :)

trieloff:
Merging this now.

PR #275, opened 2020-01-10
Fix Large Image Support

### LINKED ISSUE #274: Large files aren't delivered
This has been going on for a while, might be related to #163 and is certainly related to https://github.com/adobe/helix-pages/issues/79

trieloff:
We have been sending requests to raw.githubusercontent.com with `Authorization: token (null)` in the headers. GitHub responds to these requests with a status 404, as the token is invalid

fixes #274 and https://github.com/adobe/helix-pages/issues/79

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -1120,7 +1120,7 @@ sub hlx_bereq {
     set bereq.http.Authorization = table.lookup(secrets, "OPENWHISK_AUTH");
     # pass Github Token via X-Github-Token header
     set bereq.http.X-Github-Token = table.lookup(secrets, "GITHUB_TOKEN");
-  } elsif (req.backend == F_GitHub) {
+  } elsif (req.backend == F_GitHub && table.lookup(secrets, "GITHUB_TOKEN")) {
     # set Github backend authentication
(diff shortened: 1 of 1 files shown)


layouts/fastly/helix.vcl
```diff
@@ -1120,7 +1120,7 @@ sub hlx_bereq {
     set bereq.http.Authorization = table.lookup(secrets, "OPENWHISK_AUTH");
     # pass Github Token via X-Github-Token header
     set bereq.http.X-Github-Token = table.lookup(secrets, "GITHUB_TOKEN");
-  } elsif (req.backend == F_GitHub) {
+  } elsif (req.backend == F_GitHub && table.lookup(secrets, "GITHUB_TOKEN")) {
```
kptdobe: Why re-reading the table while you have the value in `bereq.http.X-Github-Token` ?

```suggestion
  } elsif (req.backend == F_GitHub && bereq.http.X-Github-Token) {
```

trieloff: I don't think this will work, as `bereq.http.X-Github-Token` is set in a different `elsif` branch, so it would be unset in the condition evaluation.

kptdobe: ah yes. In the PR, it is just the previous line :)

PR #278, opened 2020-01-13
Blobxy: Support for Azure Blob Proxy

trieloff:
Adds a new request type for URLs like `098af326aa856bb42ce9a21240cf73d6f64b0b45.jpg`, which will get relayed to `098af326aa856bb42ce9a21240cf73d6f64b0b45` and cached indefinitely, as the content is immutable.


trieloff:
Try

  

(with image optimization) or https://www.primordialsoup.life/hlx_098af326aa856bb42ce9a21240cf73d6f64b0b45.jpg (without)

PR #288, opened 2020-01-21
Algolia proxy

trieloff:
This PR allows us to hide Algolia queries behind the `/_query/{index}/{name}` URL and cache requests to Algolia. 

You can try by going to https://www.primordialsoup.life/_query/blog-posts/all?page=1 (this is using a clone of theblog's index)

See https://github.com/davidnuescheler/theblog/issues/44 and https://github.com/adobe/helix-index-files/issues/3

In order to be useful, we need to adjust `helix-cli` to read the Query config and send it as part of the publish request. I'll start on this as soon as this PR has been merged.

### DIFF
--- a/src/fastly/algolia.js
+++ b/src/fastly/algolia.js
@@ -13,12 +13,13 @@ function encode(expression, parameters) {
   if (!expression) {
     return '""';
   }
+  const cleanexpression = expression.replace(/\n/g, '');
   if (!parameters || parameters.length===0) {
-    return `"${encodeURIComponent(expression)}"`;
+    return `"${encodeURIComponent(cleanexpression)}"`;
   }
   return parameters.reduce((expr, param) => {
-    return '"' + expr.replace('%24%7B' + param + '%7D', `" + regsub(querystring.filter_except(req.url, "${param}"), "^.*=", "") + "`);
-  }, encodeURIComponent(expression)) + '"';
+    return '"' + expr.replace('%24%7B' + param + '%7D', `" + regsub(querystring.filter_except(req.url, "${param}"), "^[^=]*=?", "") + "`);
+  }, encodeURIComponent(cleanexpression)) + '"';
 }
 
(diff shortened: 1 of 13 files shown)


tripodsan:
> In order to be useful, we need to adjust `helix-cli` to read the Query config and send it as part of the publish request. I'll start on this as soon as this PR has been merged.

I think we need to discuss in general on how to provision the query for a helix-page projects.
/cc @dominique-pfister

src/fastly/algolia.js
```diff
@@ -0,0 +1,49 @@
...
+   + "&page=" + regsub(querystring.filter_except(req.url, "page"), "^[^=]*=?", "")
+   + "&hitsPerPage=${query.hitsPerPage}";
+}`;
+}
+
+function queryvcl(indices) {
+  function* snippets() {
+    for (let i = 0; i < indices.length; i += 1) {
+      for (let j = 0; j < indices[i].queries.length; j += 1) {
+        yield queryvclsnippet(indices[i].name, indices[i].queries[j]);
```
tripodsan: feeling adventurous :-)

trieloff:
`request.query` doesn't exist in VCL, it's `req.url.qs`, so the confusion isn't obvious to me. Would `_hlx/query` work as an alternative for you?

tripodsan:
> `request.query` doesn't exist in VCL, it's `req.url.qs`, so the confusion isn't obvious to me. Would `_hlx/query` work as an alternative for you?

I just think that reading the algolia index is not really a query, since you can't specify any query params.
but you can keep it `_query` :-)

trieloff:
You can specify query parameters. In the example, the `by-author` query does that: https://github.com/adobe/helix-publish/pull/288/files#diff-8cd3da066032408cc2072c9ec00b4a0cR25-R35

it takes the `author` URL parameter and uses it to `filter` by `author`.

PR #296, opened 2020-01-24
fix(package): include required schemas in deployment

trieloff:
Right now, running `hlx publish` fails due to the missing schemas.


trieloff:
`npx hlx publish --apiPublish https://adobeioruntime.net/api/v1/web/helix/helix-services-private/publish@ci1795` succeeds.

PR #321, opened 2020-02-13
Fallback queries

trieloff:
Basic idea: when the statically configured queries (see https://github.com/adobe/helix-publish/blob/master/test/fixtures/example-queries.vcl for an example document) do not match, fall back to calling https://github.com/adobe/helix-query-index (e.g. https://adobeioruntime.net/api/v1/web/helix/helix-services/query-index@v1/blog-posts/all?__hlx_owner=trieloff&__hlx_repo=helixdemo&__hlx_ref=3aea5fd4cd4d40f5f7c6ce3d74c6f20999903cd3) which will 307 redirect to a proper Algolia search URL. When our VCL sees that URL, it should just fetch and deliver *that*.

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -611,8 +611,25 @@ sub hlx_type_query {
   call hlx_owner;
   call hlx_repo;
 
+
+  if (req.url.path ~ "^/_query/([^/]+)\/([^/]+)$") {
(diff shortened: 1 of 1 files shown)


trieloff:
Try it here: https://www.primordialsoup.life/_query/blog-posts/new?page=1

PR #332, opened 2020-02-25
Conditions support in helix-publish

### LINKED ISSUE #155: support for strain condition language
with https://github.com/adobe/helix-shared/issues/20 we introduced a condition language that is currently not supported/used by neither helix-simulator nor helix-publish.

it would be good to improve the strain selection in the VCL logic to support the conditions.

dominique-pfister:
fix #155


dominique-pfister:
The build still fails because some expected output does not match the actual. The two types of mismatched output are:

1) `if (req.http.X-Full-URL ~ "^https://client.project-helix.io") {
      +if (req.http.Host == "client.project-helix.io") {`

This seems to amount to the same VCL condition, but I'm happy to modify the VCL returned by the Condition.

2) `-  set req.http.X-Sticky = "true";
      +  set req.http.X-Sticky = "false";`

This happens in a proxy setup: before the change, the proxy had just a `url` property, now it has a `condition` with that URL, and this automatically sets the `X-Sticky` header. Is that wrong?

trieloff:
For the stickiness change, I filed: https://github.com/adobe/helix-shared/issues/254

trieloff:
Regarding the other VCL mismatch: I think it's fine to change the test, as long as it matches semantically.

trieloff:
I'd also add a test that uses the old deprecated format, i.e. a `url` property and `condition` as a string. Depending on how we deal with it, we might have to flag this as a breaking change.

dominique-pfister:
I see that the tests mentioned (e.g. `full.yaml`) generate the following VCL from condition `url: https://developer.adobe.com/olddir`:
```
req.http.host == "developer.adobe.com" && (req.http.X-FullDirname ~ "^/olddir$" || req.http.X-FullDirname ~ "^/olddir/")
```
I'm wondering how one could derive that `olddir` should be treated as dirname and not as `req.url.path`. Is it the missing extension?

trieloff:
```yaml
url: https://developer.adobe.com/olddir
```

means:
- hostname is `developer.adobe.com`
- path is either `/olddir` or starts with `/olddir/` (so that `/olddirrigent` wouldn't match)

I believe that we are using `req.http.X-FullDirname` instead of `req.url.path` mainly to guard against scenarios where there is no extension:

https://github.com/adobe/helix-publish/blob/b0a71b21e0963555a1edea542500648b13c0e938/layouts/fastly/helix.vcl#L1088-L1097

package.json
```diff
@@ -77,7 +77,7 @@
     "@adobe/fastly-native-promises": "1.15.0",
     "@adobe/helix-epsagon": "1.1.3",
     "@adobe/helix-log": "4.5.1",
-    "@adobe/helix-shared": "5.2.2",
+    "@adobe/helix-shared": "5.3.0",
```
trieloff: This should be `v6`

Issue #372, opened 2020-03-25
url conditions are not backward compatible


tripodsan:
**Description**
using a helix-config.yaml with implizit url conditions doesn't generate the correct VCL.
Since the same still works in simulator, it is a bit confusing.

see https://github.com/adobe/project-helix.io/issues/69

### Suggestion
Drop implicit URL support completely.

dominique-pfister:
Thanks for spotting that, @tripodsan , the fallback for `strain.url` was inadvertently removed in `helix-publish`.

tripodsan:
as I said, I would drop the legacy support as we don't have any projects anymore that need it.

@trieloff WDYT?

dominique-pfister:
OK, but if we don't support `url` or `urls` in a strain in `helix-publish` anymore, couldn't we also remove that support entirely in `helix-shared`, as its version has already been bumped >= 6.0.0?

dominique-pfister:
see https://github.com/adobe/helix-shared/issues/269

PR #373, opened 2020-03-25
fix(strains): implicit url condition is not backward compatible

tripodsan:
see #372


dominique-pfister:
Thanks for spotting that, @tripodsan , the fallback for `strain.url` was inadvertently removed in `helix-publish`.

test/fixtures/urls_legacy-resolve.vcl
```diff
@@ -0,0 +1,32 @@
+# This file handles the strain resolution
+set req.http.X-Root-Path = "";
+if (req.http.Host == "www.project-helix.io" && (req.http.X-FullDirname ~ "^/client$" || req.http.X-FullDirname ~ "^/client/")) {
```
dominique-pfister: Note, that I had to uppercase `host` to make it match the generated VCL.

PR #379, opened 2020-03-30
feat(errors): new error pages for errors 403, 404, 500, 502, 503, 504

trieloff:
Updated the error page design from cloudy errors to a glitchy error page with a more subtle animation.

Screen Shot 2020-03-30 at 10 42 50

The new error pages for 403, 502, 503, and 504 aren't used in VCL yet.

### DIFF
--- a/layouts/fastly/952.html
+++ b/layouts/fastly/952.html
@@ -298,7 +298,7 @@
         <div id="wrapper">
         
         <h1 class="glitch" data-text="500">500</h1>
-        <span class="sub">imternal server error</span>
+        <span class="sub">internal server error</span>
         </div>
         </div>
   </body>
--- a/layouts/fastly/952.html
+++ b/layouts/fastly/952.html
@@ -1,252 +1,305 @@
 <!doctype html>
 <html lang="en">
   <head>
-    <title>Internal Server Error</title>
+    <title>Internal server error</title>
     <!-- Required meta tags -->
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <style>
-@import url(https://fonts.googleapis.com/css?family=opensans:500);
-body{
-                background: #111155;
-                color:#fff;
-                font-family: 'Open Sans', sans-serif;
-                max-height:700px;
-                overflow: hidden;
-            }
-            .c{
-                text-align: center;
-                display: block;
-                position: relative;
-                width:80%;
-                margin:100px auto;
-            }
-            ._404{
-                font-size: 220px;
-                position: relative;
-                display: inline-block;
-                z-index: 2;
-                height: 250px;
-                letter-spacing: 15px;
-            }
-            ._1{
-                text-align:center;
-                display:block;
-                position:relative;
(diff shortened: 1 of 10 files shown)


layouts/fastly/952.html
```diff
@@ -1,252 +1,305 @@
...
-      <hr>
-      <div class='_1'>INTERNAL</div>
-      <div class='_1'>SERVER ERROR</div>
-      <code class='btn'>There was an error processing your request.</code>
-  </div>
+    <div id="app">
+        <div id="wrapper">
+        
+        <h1 class="glitch" data-text="500">500</h1>
+        <span class="sub">imternal server error</span>
```
kptdobe:
```suggestion
        <span class="sub">internal server error</span>
```

layouts/fastly/953.html
```diff
@@ -1,252 +1,305 @@
...
-      </body>
-    </html>
\ No newline at end of file
+  </head>
+  <body>
+    <div id="app">
+        <div id="wrapper">
+        
+        <h1 class="glitch" data-text="500">500</h1>
+        <span class="sub">imternal server error</span>
```
kptdobe:
```suggestion
        <span class="sub">internal server error</span>
```

layouts/fastly/generic-error.html
```diff
@@ -1,252 +1,305 @@
...
-      </body>
-    </html>
\ No newline at end of file
+  </head>
+  <body>
+    <div id="app">
+        <div id="wrapper">
+        
+        <h1 class="glitch" data-text="500">500</h1>
+        <span class="sub">imternal server error</span>
```
kptdobe:
```suggestion
        <span class="sub">internal server error</span>
```

kptdobe:
I did not look at all of them, typo might be everywhere.

kptdobe:
Question: it seems that most of error pages are the same... why do we need 5 or 6 times to copy the same 500 error page with a different name ?
If we make no distinction between 500, 951, 952... and generic error, why we simply not just change the code to load the same single error page... ? The day we need a different error page, we can still adjust the code to load a different version.

kptdobe:
Also, we need to be careful with the 1MB limit for the custom VCL file size... It did not complain yet but we might approach this limit... https://docs.fastly.com/en/guides/resource-limits#vcl-and-configuration-limits

trieloff:
Each error page is 77 kb, so we still have a bit to go. I wanted to include the background image to increase self-sufficiency, but we can probably downscale and reduce the quality a bit more if push comes to shove.

kptdobe:
Approved and tested: 

https://alex.helix-demo.xyz/.bla.html - 403
 https://alex.helix-demo.xyz/e.404.html - 404
 https://alex.helix-demo.xyz/e.500.html - 500
 https://alex.helix-demo.xyz/e.502.html - 500
 https://alex.helix-demo.xyz/e.503.html - 503
 https://alex.helix-demo.xyz/e.504.html - 504

kptdobe:
I allowed myself to merge so I can nove forward with https://github.com/adobe/helix-publish/pull/382

Issue #167, opened 2018-12-07
hlx publish should validate the action path before publishing


trieloff:
I've spend the better part of the morning trying to fix a 404 which I could ultimately resolve to a typo in a config file. (At least we got a shiny new `X-Trace` header out of it, which will be useful for further VCL debugging).

We can prevent this by having `hlx publish` validate that the provided OpenWhisk functions actually exist before publishing.

MarquiseRosier:
@trieloff  @tripodsan 

for clarity; publish will need openwhisk credentials if we plan to make it verify actions deployed for strains, as these are deployed under authors namespace

MarquiseRosier:
I was thinking this would best be done in the cli remotePublish.cmd.js file. we can avoid whisk credentials having to be passed into publish service, and it'll be more tolerable, to deal with whatever additional runtime is accrued from filtering whisk to check actions are deployed before a publish, on the cli!

tripodsan:
there are 2 possible cases:
1. the action is public, so it can be accessed via the web url. this is the default case
2. the action is secured, in this case the edge needs to know the wsk secret anyways, which is usually stored in the edge directory during publish. hence the publish action needs to know the secret aswell.

IMO, we should check the action coordinates in helix-publish

MarquiseRosier:
@tripodsan 

For this task; we have to just send a request to the adobeioruntime.net/api/v1/namespaces/{NAMESPACE}/packages/{PACKAGE}

this requires authentication though is there any way to not have to pass these credentials to the service through params? I was thinking to use the edge dictionary 'secrets' like you mentioned @tripodsan but then it's write_only.

tripodsan:
right, I thought that we pass the `wsk_auth` already to the `helix-publish` action, but this is still done int the cli: https://github.com/adobe/helix-cli/blob/7de7b326a19309a3263155e23b132b5bc6a792ae/src/remotepublish.cmd.js#L328

IMO, this is not consistent with the rest of the work that helix-publish does, as it also updates other edge dictionaries. 

@trieloff should we pass the `wsk_auth` to helix-publish and do the entire setup in the action?

trieloff:
Yes, that's probably the right way of doing it.

MarquiseRosier:
Additionally; we have to pass other things like the GitConfig, fastly credentials, and wsk credentials; if we plan to move the publish functionality out of the CLI.

PR #394, opened 2020-04-07
Package check

MarquiseRosier:

## Related Issues
#167 

Thanks for contributing!


package.json
```diff
@@ -85,8 +87,15 @@
     "@adobe/helix-status": "7.1.2",
     "@adobe/openwhisk-action-logger": "2.2.0",
     "@adobe/openwhisk-action-utils": "4.2.2",
+    "cz-conventional-changelog": "^3.1.0",
     "fs-extra": "9.0.0",
     "glob-to-regexp": "0.4.1",
+    "openwhisk": "^3.21.1",
```
tripodsan, line 93: why is this a dev dependency?

src/check-pkgs.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param namespace {string} openwhisk namespace under which actions exist
+ */
+function check_pkgs(auth, host, namespace, config) {
+  return new Promise((resolve, reject) => {
+    const ow = openwhisk({
+      api_key: auth,
+      apihost: host,
+      namespace,
+    });
+    ow.packages.list().then((data) => {
```
tripodsan, line 26: why use promises and not async/await ?

src/check-pkgs.js
```diff
@@ -0,0 +1,43 @@
...
+      const package_list = data.reduce((prev, curr) => {
+        prev[curr.name] = true;
+        return prev;
+      }, {});
+
+      config.strains.forEach((strain) => {
+        if (strain.package && !(strain.package in package_list)) {
+          return reject(Error(`config package for the strain: ${strain.name} not deployed`));
```
tripodsan, line 34:
```suggestion
          return reject(Error(`action package ${strain.package} for the strain: ${strain.name} not deployed`));
```

src/publish.js
```diff
@@ -29,7 +29,7 @@ const redirects = require('./fastly/redirects');
  * @param {object} iconfig the IndexConfig from helix-shared
  * @param {string} algoliaappid Algolia App ID (not secret)
  */
-async function publish(configuration, service, token, version, vclOverrides = {}, dispatchVersion = 'v3', log = console, iconfig, algoliaappid) {
+async function publish(configuration, service, token, version, vclOverrides = {}, dispatchVersion = 'v3', log = console, iconfig, algoliaappid, wskAuth, wskHost, wskNamespace) {
```
tripodsan, line 32: this gets longer and longer....maybe switch to a different style. eg:
- pass a `options` object
- use a builder pattern

(but this can be addresses in a different PR)

test/fixtures/recordings/Integration-Test_519117731/Test-publish-function-locally_1268185746/recording.har
```diff
@@ -4930,6 +4930,105 @@
...
+            {
+              "name": "host",
+              "value": "fake_host"
+            }
+          ],
+          "headersSize": 347,
+          "httpVersion": "HTTP/1.1",
+          "method": "GET",
+          "queryString": [],
+          "url": "https://fake_host/api/v1/namespaces/fake_ns/packages"
```
tripodsan, line 4966: how did you record this?

src/index.js
```diff
@@ -27,6 +27,9 @@ async function publishConfig(params) {
     params.__ow_logger,
     params.indexconfig,
     params.algoliaappid,
+    params.wskAuth,
+    params.wskHost,
```
tripodsan, line 31: this should default to `process.env.__OW_API_HOST`

src/check-pkgs.js
MarquiseRosier: I wanted to have it resolved by the call to Promise.all

test/fixtures/recordings/Integration-Test_519117731/Test-publish-function-locally_1268185746/recording.har
MarquiseRosier: I first recorded call to openwhisk with my credentials; and then put in all the fake stuff in the recordings

src/index.js
MarquiseRosier: ^ Right! Good call, i'll fix this asap tomorrow morning

src/check-pkgs.js
MarquiseRosier: I don't think it'll lead to a race condition, I checked and check-pkgs always resolves first. so it'll cause the entire Promise.all to fail before anything is written to the VCL/dictionaries, if you think it's not safe for racing reasons, I can move it. and use the await notation as you advise to ensure it executes first.

tripodsan: I don't understand...`Promise.all()` doesn't guarantee the order of when the promises are resolved. however, this is not the point.

the `new Promise()` here is returned.... this is the same as writing:

```js
async function check_pkgs(auth, host, namespace, config) {
...
```

test/fixtures/recordings/Integration-Test_519117731/Test-publish-function-locally_1268185746/recording.har
tripodsan: right...but why change the host ?

MarquiseRosier: right, that was probably overkill. Would it be any trouble to leave it? The authorization is fake

tripodsan: sure...just keep it...I was just curious.

src/check-pkgs.js
MarquiseRosier: arghhh duhhh it always returns a promise! I will fix this.

src/check-pkgs.js
```diff
@@ -0,0 +1,86 @@
...
+async function getPackageList(auth, host, namespace) {
+  let packageList;
+  const ow = openwhisk({
+    api_key: auth,
+    apihost: host,
+    namespace,
+  });
+
+  try {
+    packageList = await ow.packages.list();
```
tripodsan, line 50:
```suggestion
    const packageList = await ow.packages.list();
```

src/check-pkgs.js
```diff
@@ -0,0 +1,86 @@
...
+    return packageList.reduce((prev, curr) => {
+      // eslint-disable-next-line no-param-reassign
+      prev[curr.name] = true;
+      return prev;
+    }, {});
+  } catch (e) {
+    throw new WhiskError('whisk failed to obtain package list');
```
tripodsan, line 57: maybe log the error?

src/check-pkgs.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+}
+
+/**
+ * @param auth {string} openwhisk authentication key
+ * @param host {string} openwhisk api host
+ * @param namespace {string} openwhisk namespace under which actions exist
+ * @param config {object} a Helix Configuration object
+ *
+ * This function checks that packages defined for
```
tripodsan, line 67: convention is to have the description before the params.

src/check-pkgs.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+}
+
+/**
+ *
+ * @param auth {string} openwhisk authentication key
+ * @param host {string} openwhisk api host
+ * @param namespace {string} openwhisk namespace under which actions exist
+ *
+ * This function gets list of packages deployed to OpenWhisk
```
tripodsan, line 39: convention is to have the description before the params.

src/check-pkgs.js
MarquiseRosier: True, I left it because the error will get logged in the publish function. 

https://github.com/adobe/helix-publish/blob/331fb33b81761854e60950651189b25abd8e671c/src/publish.js#L98

src/check-pkgs.js
MarquiseRosier: True, I left it because the error will get logged in the publish function. 

https://github.com/adobe/helix-publish/blob/331fb33b81761854e60950651189b25abd8e671c/src/publish.js#L96-L105

src/index.js
MarquiseRosier: I think this will happen in the publish command in helix-cli!

MarquiseRosier: https://github.com/adobe/helix-cli/blob/318770004ea81dd88bc825b73fc124343f24efd0/src/remotepublish.cmd.js#L285-L326

src/check-pkgs.js
tripodsan: I mean, log the reason for the error... the `e`

package.json
trieloff: Just a guess: because it's pre-installed on the container but needed during testing. Making it a regular dependency would be cleaner, though.

src/check-pkgs.js
tripodsan: @MarquiseRosier so...will you add the error logging here?

package.json
MarquiseRosier: I think this is outdated now

tripodsan: it's about `openwhisk`

test/fixtures/recordings/Integration-Test_519117731/Test-publish-function-locally_1268185746/recording.har
```diff
@@ -4930,6 +4930,105 @@
...
+              "name": "accept",
+              "value": "application/json"
+            },
+            {
+              "name": "user-agent",
+              "value": "openwhisk-client-js"
+            },
+            {
+              "name": "authorization",
+              "value": "Basic ZmFrZV9hdXRo"
```
trieloff, line 4951: This reads `fake_auth`. 😥

src/index.js
```diff
@@ -27,6 +27,9 @@ async function publishConfig(params) {
...
+    params.wskNamespace,
```
tripodsan, line 32: please add fallback to:

- `process.env.__OW_API_KEY`
- `process.env.__OW_NAMESPACE`
- `process.env.__OW_API_HOST`

then, we don't need to pass the creds in `helix-cli`. see https://github.com/adobe/helix-cli/issues/1378#issuecomment-616878803

test/fixtures/recordings/Integration-Test_519117731/Test-publish-function-locally_1268185746/recording.har
MarquiseRosier: haha disappointed hacker

.releaserc.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
     '@semantic-release/commit-analyzer',
     '@semantic-release/release-notes-generator',
     ['@semantic-release/changelog', {
-      changelogFile: 'CHANGELOG.md',
+      chadangelogFile: 'CHANGELOG.md',
```
tripodsan, line 6: ????

MarquiseRosier:
sweet thanks a ton @tripodsan!

PR #425, opened 2020-04-28
feat(publish): fix package check

### LINKED ISSUE #423: Re-Enable (working) package check
@MarquiseRosier @trieloff can you create a follow up issue to enable it again and fix it in helix-cli ?

_Originally posted by @tripodsan in https://github.com/adobe/helix-publish/issues/416#issuecomment-619974774_

MarquiseRosier:
fix package check

fix #423

- [x] make sure to link the related issues in this description
- [x] when merging / squashing, make sure the fixed issue references are visible in the commits, for easy compilation of release notes

## Related Issues

Thanks for contributing!


MarquiseRosier:
You can check that it works by using 
```
hlx publish --wsk-namespace blahblah --fastly-serviceid blahblah --api-publish --api-publish https://adobeioruntime.net/api/v1/web/mrosier/helix-services/publish@ci2425
```

test/fixtures/demo.yaml
```diff
@@ -7,7 +7,7 @@ strains:
     content: https://github.com/trieloff/helix-demo.git#master
     static: https://github.com/trieloff/helix-demo.git/htdocs#master
     code: ssh://git@github.com/trieloff/helix-demo.git#master
-    package: trieloff/adeb2be318b4997c8254862e8815d8c62fe4de49
+    package: mrosier/adeb2be318b4997c8254862e8815d8c62fe4de49
```
tripodsan, line 10: what is try? the package should not contain the namespace.

@trieloff ?

MarquiseRosier: ahhh that's odd. That's actually the only reason it's failing! :D is because the packages are named with the namespace in helix-configs!

src/check-pkgs.js
```diff
@@ -71,13 +70,17 @@ async function checkPkgs(auth, host, namespace, config, log = console) {
   const packages = await getPackageList(auth, host, namespace, log);
 
   config.strains.forEach((strain) => {
-    if (strain.package && !(strain.package in packages)) {
-      log.error(`action package for the following strain: << ${strain.name} >> not deployed`);
+    const package = new RegExp(`(?<=${namespace}/).*`).exec(strain.package);
```
tripodsan, line 73: this is not very efficient. you create the regexp for each strain, although the namespace never changes. also, using regexp looks a bit complicated...

also, what happens if the namespace in the `strain.package` doesn't match the namespace provided?

what about something along the lines of
```
const segments = strain.package.split('/');
if (segments.length > 1) {
  if (segments[0] !== namespace) throw error('packages with different namespaces are not suported');
  segments.shift();
}
const package = segments[0];

Comment (MarquiseRosier) [reply src/check-pkgs.js:73]: Ok! I can do that! 

I'm confused about your second part. If the namespace in the strain.package doesn't match it'll be not Found, which seems like what should happen

Comment (tripodsan) [reply src/check-pkgs.js:73]: yes, but the error is different. if the package cannot be found, then you didn't deploy the package via `hlx deploy`. but if the namespace is wrong (and if we gonna support this), then there is a more fundamental problem, that you try to publish something and having a wrong wsk namespace setup.

however, I think this is all a bit wrong :-)

since every package that we `hlx deploy` will have a `hlx--static` action, eg:

```
$ hlxctl info /helix-pages/15c18adcdfe346441d5b2f422c1e44cfec10b71d
Actions in package /helix-pages/15c18adcdfe346441d5b2f422c1e44cfec10b71d
  cgi-bin-sitemap  nodejs:10
  hlx--static      sequence -> /helix-pages/helix-services/static@v1
  html             nodejs:10
  idx_json         nodejs:10
  plain_html       nodejs:10
```

it would be easier to check if that action exists, since it is anyways needed for successful operation.

so the better check you be to invoke ` https://adobeioruntime.net/api/v1/web/${namespace}/${package}/hlx--static/_status_check/healthcheck.json`. this tests both, if the package is present and the static action.

eg:
```
curl -sD- https://adobeioruntime.net/api/v1/web/helix-pages/15c18adcdfe346441d5b2f422c1e44cfec10b71d/hlx--static/_status_check/healthcheck.json
```

Comment (MarquiseRosier) [reply src/check-pkgs.js:73]: ^^ Thank you for this info!

File: test/fixtures/demo.yaml (hunk 1 above)
Comment (trieloff) [reply test/fixtures/demo.yaml:10]: I see a number of use cases for different namespaces in one config:

- you have different developers in one team, but don't want to share open whisk credentials
- you want to use another users actions as a "theme", similar to what @MarquiseRosier alluded to

**Comment by trieloff** [comment]
> so the better check could be to invoke `https://adobeioruntime.net/api/v1/web/${namespace}/${package}/hlx--static/_status_check/healthcheck.json`. this tests both, if the package is present and the static action.
>
> The advantage is also, that the openwhisk credentials are not needed and that it works on all namespaces.

I like this approach much better.

For the scenario that the Runtime actions *are* password protected, we should first try to access the health check without authentication and if that fails (with a 401 or 403) then with authentication.

Authentication can be provided out of the environment (which would require the change in helix-cli to use the user's namespace for helix-publish) or as explicit parameters.

**Comment by tripodsan** [comment]
> For the scenario that the Runtime actions _are_ password protected, we should first try to access the health check without authentication and if that fails (with a 401 or 403) then with authentication.
> 
> Authentication can be provided out of the environment (which would require the change in helix-cli to use the user's namespace for helix-publish) or as explicit parameters.

ok. but let's start w/o the private actions... currently we don't have a way to specify this anyways during `hlx deploy`.

File: src/check-pkgs.js [hunk 3]
```diff
@@ -10,77 +10,31 @@
...
-      // eslint-disable-next-line no-param-reassign
-      prev[curr.name] = true;
-      return prev;
-    }, {});
-  } catch (e) {
-    log.error(e);
-    throw new WhiskError('whisk failed to obtain package list');
-  }
-}
+const fetch = require('@adobe/helix-fetch');
```
trieloff, line 13: Why not

```suggestion
const { fetch } = require('@adobe/helix-fetch');
```

src/check-pkgs.js
```diff
@@ -10,77 +10,31 @@
...
-    const package = new RegExp(`(?<=${namespace}/).*`).exec(strain.package);
-    if (package && !(package[0] in packages)) {
-      throw new PackageNotFoundError(
-        `action package for the following strain: << ${strain.name} >> not deployed`,
-      );
+async function checkPkgs(config) {
+  const checks = config.strains.getRuntimeStrains().reduce((acc, curr) => {
+    if (curr.package) {
+      acc.push(new Promise((resolve, reject) => {
+        fetch.fetch(`https://adobeioruntime.net/api/v1/web/${curr.package}/hlx--static/_status_check/healthcheck.json`)
```
tripodsan, line 25: what if the `curr.package` doesn't include the namespace?

tripodsan: why not use `await` ?

src/check-pkgs.js
```diff
@@ -10,77 +10,31 @@
...
+          .then((res) => {
+            if (!res.ok) {
+              reject(new Error(`the following health check failed: ${res.url}`));
+            }
+            resolve();
```
tripodsan, line 30: also remember to consume the response, otherwise fetch might hang the process.

src/check-pkgs.js
MarquiseRosier: @tripodsan I don't think this can happen however, because of the way the deploy command works right? 

I didn't use await because I wanted to run the requests concurrently, and not wait for them, was this wrong?

src/check-pkgs.js
```diff
@@ -10,74 +10,31 @@
...
-    if (strain.package && !(strain.package in packages)) {
-      log.error(`action package for the following strain: << ${strain.name} >> not deployed`);
+async function checkPkgs(config) {
+  const checks = config.strains.getRuntimeStrains().reduce((acc, curr) => {
+    if (curr.package) {
+      acc.push(new Promise((resolve, reject) => {
+        fetch(`https://adobeioruntime.net/api/v1/web/${curr.package}/hlx--static/_status_check/healthcheck.json`)
+          .then((res) => {
+            if (!res.ok) {
+              reject(new Error(`the following health check failed: ${res.url}`));
```
tripodsan, line 28: also consume the response on error...otherwise helix-fetch will hang the action.

src/check-pkgs.js
```diff
@@ -10,74 +10,31 @@
...
-  const packages = await getPackageList(auth, host, namespace, log);
-
-  config.strains.forEach((strain) => {
-    if (strain.package && !(strain.package in packages)) {
-      log.error(`action package for the following strain: << ${strain.name} >> not deployed`);
+async function checkPkgs(config) {
+  const checks = config.strains.getRuntimeStrains().reduce((acc, curr) => {
+    if (curr.package) {
+      acc.push(new Promise((resolve, reject) => {
+        fetch(`https://adobeioruntime.net/api/v1/web/${curr.package}/hlx--static/_status_check/healthcheck.json`)
```
tripodsan, line 25: I would move this in a separate function and als use async/await....

test/check-pkg.test.js
```diff
@@ -23,7 +23,7 @@ const checkPkgs = require('../src/check-pkgs');
 describe('test check_pkgs', () => {
   setupPolly({
     recordFailedRequests: true,
-    recordIfMissing: false,
+    recordIfMissing: true,
```
rofe, line 26: I would set it back to false

package.json
```diff
@@ -82,14 +82,14 @@
   "dependencies": {
     "@adobe/fastly-native-promises": "1.16.0",
     "@adobe/helix-epsagon": "1.2.0",
+    "@adobe/helix-fetch": "^1.6.0",
```
rofe, line 85: Renovate will send a PR next in order to pin this dependency 😉

test/fixtures/demo.yaml
rofe: Nitpicking, but maybe we should switch to a dummy namespace in the fixtures and recordings.

src/check-pkgs.js
```diff
@@ -10,74 +10,43 @@
...
- * @param host {string} openwhisk api host
- * @param namespace {string} openwhisk namespace under which actions exist
- */
-async function getPackageList(auth, host, namespace, log) {
-  const ow = openwhisk({
-    api_key: auth,
-    apihost: host,
-    namespace,
-  });
+async function getPromises(strains, log) {
```
tripodsan, line 21: not a very good function name :-) what about check package ?

src/check-pkgs.js
```diff
@@ -10,74 +10,43 @@
...
+  const checks = strains.reduce((acc, curr) => {
+    if (curr.package) {
+      acc.push(fetch(`https://adobeioruntime.net/api/v1/web/${curr.package}/hlx--static/_status_check/healthcheck.json`)
+        .then(async (result) => {
```
tripodsan, line 25: still a `then`... why not async?

src/check-pkgs.js
```diff
@@ -10,74 +10,43 @@
...
- */
-async function getPackageList(auth, host, namespace, log) {
-  const ow = openwhisk({
-    api_key: auth,
-    apihost: host,
-    namespace,
-  });
+async function getPromises(strains, log) {
+  const checks = strains.reduce((acc, curr) => {
+    if (curr.package) {
```
tripodsan, line 23: this check can be outside...

src/check-pkgs.js
```diff
@@ -10,74 +10,43 @@
...
-  const packages = await getPackageList(auth, host, namespace, log);
-
-  config.strains.forEach((strain) => {
-    if (strain.package && !(strain.package in packages)) {
-      log.error(`action package for the following strain: << ${strain.name} >> not deployed`);
-    }
-  });
+async function checkPkgs(config, log = console) {
+  const strains = config.strains.getRuntimeStrains();
+  return Promise.all(await getPromises(strains, log));
```
tripodsan, line 49: I would do something like:

```
const tasks = strains
  .filter((s) => !!s.package)
  .map(checkPackage)
```

MarquiseRosier: Thanks for these tips Tobi! :)

tripodsan:
@MarquiseRosier still working on this?

tripodsan:
@rofe:

> Nitpicking, but maybe we should switch to a dummy namespace in the fixtures and recordings.

it's a bit difficult to do a recording using a fake namespace. since the recording-matching contains the URL, i.e. the namespace.

can you approve, so we can finally merge this?

PR #446, opened 2020-05-18
Tracing VCL execution with Epsagon

trieloff:
This (WIP) PR adds the ability to inject additional log lines in VCL code. The log statements will be in a JSON format that can be understood by Epsagon and trace:

- entering a subroutine
- leaving a subroutine normally
- calling a different subroutine
- going into a different state in the state machine

There are a few more things that need to be logged:

- [ ] service ID
- [ ] current time
- [ ] start time


ranrib:
Something looks weird with the IDs that we are getting - for example request-id: `51f047ed-3d28-a9e-9df3-1ae1d1418303` - the middle part (`a9e`) is only 3 letters long and should be 4.
Do you have any idea about it?

trieloff:
https://github.com/adobe/helix-publish/blob/a7a8446ca12e17454d5aef33c6592b1985c04006/layouts/fastly/helix.vcl#L137

Yes, we only create three characters in the middle. It's easy to fix.

trieloff:
@ranrib @stefan-guggisberg @kptdobe @tripodsan If we are ok with keeping the log payload format open for iteration (something that should only affect @ranrib's side of things), I'd consider this ready to be merged.

src/vcl/tracer.js
```diff
@@ -0,0 +1,142 @@
...
+// list of values to ignore every time
+const ignore = [
+  'req.http.X-PreFetch-Miss',
+  'req.http.X-PreFetch-Pass',
+  'req.http.X-Trace',
+  'beresp.http.X-PreFetch-Miss',
+  'beresp.http.X-PreFetch-Pass',
+];
+
+// list of values to include every time (rember that only req is readable at any time)
```
stefan-guggisberg, line 27: remember

PR #459, opened 2020-06-03
feat(vcl): add hlx-forwarded-host header

### LINKED ISSUE #451: Set hlx-forwarded-host header from x-forwarded-host
the canonical URL for everything that doesn't have a `x-cdn-url` header set is still pointing to `*.hlx.page` which is very undesirable from an SEO standpoint, meaning that helix pages can only be found in the `*.hlx.page` domain... 

i appreciate the issue we have with the `x-forwarded-host` we have with the implementation in runtime, so i think in the meantime we should set the `x-cdn-url` header with the value from `x-forwarded-host` if present in the inner CDN. this allows us to shield any other outer cdn from the implementation issues of `x-forwarded-host` in runtime.

_Originally posted by @davidnuescheler in https://github.com/adobe/helix-pages/issues/284#issuecomment-634779047_

trieloff:
fixes #451

## Related Issues

Thanks for contributing!

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -1457,6 +1457,10 @@ sub hlx_bereq {
     }
     if (req.backend == F_AdobeRuntime) {
       set bereq.http.Host = "adobeioruntime.net";
+      // Adobe I/O Runtime overrides the Host and Forwarded-Host
+      // headers, so we create a new one that Runtime won't
+      // override
+      set bereq.http.hlx-forwarded-host = req.http.X-Orig-Host;
     } elsif (req.backend == F_GitHub) {
       set bereq.http.Host = "raw.githubusercontent.com";
     } elseif (req.backend == F_AzureBlobs) {


rofe:
@trieloff what's the reason for the changed source header (`x-orig-host` vs `x-forwarded-host`)? I thought you wanted to use the latter? In any case, it might be beneficial to have a test to make sure we won't break it in the future.

trieloff:
`x-orig-host` is used internally and should always be consistent. But, we make sure not to share it with any origin. 

`x-forwarded-host` would be a good header to send to Adobe I/O Runtime, except that Runtime overwrites it with `adobeioruntime.net`

`hlx-forwarded-host` is the header we are now sharing selectively (only when the backend is Runtime)

rofe:
Thanks for the clarification!

> `hlx-forwarded-host` is the header we are now sharing selectively (only when the backend is Runtime)

I assume this is also the case for cgi-bin actions, yes?

PR #460, opened 2020-06-03
fix(vcl): serve 429 error page when no response body exists

### LINKED ISSUE #461: Fastly returns 500 on 429 received from origin
**Description**
We've experienced cases where the origin (`dispatch` action) returned status `429` (too many requests, I/O runtime limit) and Fastly returned status `500`. 

**To Reproduce**
See https://helix.coralogix.com/#/query/logs?id=poqVh1PgS5Dx4SkSCNWVmW for an example. 

**Expected behavior**
Fastly should return status `503` when the origin returns `429`.

trieloff:

## Related Issues

Thanks for contributing!

Fixes #461


trieloff:
I think we should leave the distinction.

PR #497, opened 2020-06-17
feat(vcl): block requests by URL pattern

trieloff:
Test cases include known bad URLs (thanks @asanso) and the top 1000 paths on helix-pages as good URLs


trieloff:
@rofe FYI

PR #509, opened 2020-06-25
fix(vcl): suppress Age header

### LINKED ISSUE #506: browser cache being disabled because of leaked age header
**Description**
As stated by @davidnuescheler  

@trieloff and others, i just noticed that the .json requests for the query index don't get cached in my browser anymore... after looking into it, i noticed that this is probably because we send the following:
age: 13608
cache-control: max-age=7200

after flushing the inner cdn to rejuvenate the cache, the browser cache works again, with the following headers...
age: 31
cache-control: max-age=7200
New

i decided that this is an outer CDN issue, that i assigned to @trieloff https://github.com/adobe/theblog/issues/253 feel free to reassign if there is a better person to quickly fix this.

I think the fix is just to unset the resp.http.Age if an X-Debug header isn't sent in the request, then we can just unset the resp.http.Age? Forgive me if i'm assuming some knowledge here; 

https://github.com/adobe/helix-publish/blob/a8265b7e679975eea4a82a4eb1e3f769ccad9c6d/layouts/fastly/helix.vcl#L116-L130

or we can unset it here:

https://github.com/adobe/helix-publish/blob/a8265b7e679975eea4a82a4eb1e3f769ccad9c6d/layouts/fastly/helix.vcl#L372-L403

**To Reproduce**
send a curl request to a hlx.page site and look at the headers!

trieloff:
fixes #506

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -1730,7 +1730,11 @@ sub vcl_deliver {
     unset resp.http.x-xss-protection;
   } else {
     set resp.http.X-Trace = req.http.X-Trace;
+    set resp.http.X-Age = resp.http.Age;
   }
+
+  // see https://github.com/adobe/helix-publish/issues/506
+  unset resp.http.Age;
 
   set resp.http.X-Frame-Options = "DENY";
 


stefan-guggisberg:
Hmm.. I initially thought about the same solution (unconditionally strip `Age` headers on CDN delivery. However, after some researching and testing with Fastly Fiddle I am not sure if this doesn't potentially cause other problems. 

Fastly updates/sets the `Age` header on delivery (verified with Fiddle).

trieloff:
This is the solution suggested by Fastly.

PR #555, opened 2020-07-30
fix(backends): increase Fastly timeouts

trieloff:
as suggested by @tysonnorris in https://adobe.slack.com/archives/C68UJPDM5/p1596033493473200?thread_ts=1595840508.428700&cid=C68UJPDM5

This could reduce the amount of 503s we are seeing.


stefan-guggisberg:
60s connect timeout seems excessive. I'd start with 5s. If we're still seeing connect timeouts then something's truly broken.

trieloff:
@stefan-guggisberg I've updated to 5 seconds.

stefan-guggisberg:
@trieloff You've also decreased the `between-bytes-timeout` to 5s (used to be 10s, the default).  I would leave that at the default value.

trieloff:
Done

PR #557, opened 2020-07-31
feat(blobs): include SAS token for fetching blobs if configured

### LINKED ISSUE #556: Add possibility to include SAS token for hlx_blob backend requests
the `hlx_blob` handling in the VCL uses an azure blob storage as backend. currently it needs to be configured to allow anonymous access to read the blobs. for extra security it might be needed to change their access level to private.
the backend request would need to include an SAS token, which needs to be configurable during helix publish.

**Describe the solution you'd like**
- add a `AZURE_BLOB_SAS_RO` parameter
- add it to a dictionary
- when present, append it to the backend request

**Additional context**

example:

private blob storage url w/o SAS:
- https://hlx.blob.core.windows.net/private-test/parrots.png

private blob storage url with SAS
- https://hlx.blob.core.windows.net/private-test/parrots.png?si=ro-blobs&sv=2019-12-12&sr=c&sig=IXwCW%2BQapoG1wvPrENX3pagBA4W64V6TlGnFjTz8%2FdY%3D

tripodsan:
fixes #556

@trieloff  Not sure if/how this should work.

- I don't understand how the `secrets` dictionary is updated
- does the `+` work, if the item is not in the dictionary? would it just return an empty string?

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -630,16 +630,18 @@ sub hlx_type_blob {
 
   declare local var.sha STRING;
   declare local var.ext STRING;
+  declare local var.sas STRING;
 
   if (req.url.path ~ "^/hlx_(([0-9a-f]){40}).(jpg|jpeg|png|webp|gif)$") {
     set var.sha = re.group.1;
     set var.ext = req.url.ext;
+    set var.sas = table.lookup(secrets, "AZURE_BLOB_SAS_RO");
 
     # request image optimization
     set req.http.X-Fastly-Imageopto-Api = "fastly";
   }
 
-  set req.http.X-Backend-URL = "/external/" + var.sha;
+  set req.http.X-Backend-URL = "/external/" + var.sha + var.sas;
 }
 
 sub hlx_type_query {
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -635,7 +635,7 @@ sub hlx_type_blob {
   if (req.url.path ~ "^/hlx_(([0-9a-f]){40}).(jpg|jpeg|png|webp|gif)$") {
     set var.sha = re.group.1;
     set var.ext = req.url.ext;
-    set var.sas = table.lookup(secrets, "AZURE_BLOB_SAS_RO");
+    set var.sas = table.lookup(secrets, "AZURE_BLOB_SAS_RO", "");
 
     # request image optimization
     set req.http.X-Fastly-Imageopto-Api = "fastly";


layouts/fastly/helix.vcl
```diff
@@ -630,16 +630,18 @@ sub hlx_type_blob {
 
   declare local var.sha STRING;
   declare local var.ext STRING;
+  declare local var.sas STRING;
 
   if (req.url.path ~ "^/hlx_(([0-9a-f]){40}).(jpg|jpeg|png|webp|gif)$") {
     set var.sha = re.group.1;
     set var.ext = req.url.ext;
+    set var.sas = table.lookup(secrets, "AZURE_BLOB_SAS_RO");
```
tripodsan, line 638: what happens if the item is not in the dictionary?

trieloff:
```suggestion
    set var.sas = table.lookup(secrets, "AZURE_BLOB_SAS_RO", "");
```

https://developer.fastly.com/reference/vcl/functions/table/table-lookup/

The third argument is the default return value.

trieloff:
> I don't understand how the `secrets` dictionary is updated

So far we only update it in helix-cli

https://github.com/adobe/helix-cli/blob/0930ece05758fa1f41df343c480be39ae88bfaa8/src/remotepublish.cmd.js#L382-L412

How sensitive is the SAS? If it's insensitive, we can provide the value as a default parameter to `helix-publish` and set it here. Customers will still be able to read the value with a bit of custom VCL.

tripodsan:
> How sensitive is the SAS? If it's insensitive, we can provide the value as a default parameter to `helix-publish` and set it here. 

The whole idea is to enable blocking anonymous access to the blobs. if the SAS token is public, it defies the purpose.
I expect that eventually, we will have a blob store separate for each customer. how this will exactly work with the services that write the blobs is not very clear yet.

> Customers will still be able to read the value with a bit of custom VCL.

yes. but they also set the value when running `hlx publish` :-) so they already know it :-)

trieloff:
> yes. but they also set the value when running `hlx publish` :-) so they already know it :-)

Not if `helix-publish` (service) sets the value instead of `hlx publish` (CLI). This keeps it at least somewhat hidden and you need at least a Fastly account to take a peek. 

We could generate one SAS for each service config, similar to how `helix-logging` generates service accounts.

trieloff:
But that means that only helix-pages will be able to use word2md and gdocs2md because the blob bucket is hard configured and only we know the SAS.

tripodsan:
> But that means that only helix-pages will be able to use word2md and gdocs2md because the blob bucket is hard configured and only we know the SAS.

- yes, but a non helix-pages customer will have its own fastly service and can configure a different block storage (to be implemented)....which will be more complicated anyways, since they also would need a different SAS/block configuration for word2md etc...

trieloff:
Yep.

Issue #576, opened 2020-08-19
make blocklist regex easier to manage


MarquiseRosier:
**Is your feature request related to a problem? Please describe.**
separate blocklist regex from vcl code; as it is too long

PR #577, opened 2020-08-19
fix(vcl.js): make regex management easier

MarquiseRosier:
separates blocklist from vcl to make managing it simpler

## Related Issues
#576 

Thanks for contributing!


src/fastly/include-util.js
```diff
@@ -13,6 +13,14 @@
 const fs = require('fs-extra');
 const path = require('path');
 
+function regex(content, filePath) {
+  return content.replace(/\{"regex:block.rgx"\}/g, () => {
+    const name = path.resolve(filePath);
+    const c = fs.readFileSync(name).toString().split('\n').join('');
```
tripodsan, line 19: - why sync ?
- why not `readFile(name, 'utf-8')` (no toString() needed)

src/fastly/include-util.js
```diff
@@ -13,6 +13,14 @@
 const fs = require('fs-extra');
 const path = require('path');
 
+function regex(content, filePath) {
+  return content.replace(/\{"regex:block.rgx"\}/g, () => {
+    const name = path.resolve(filePath);
```
tripodsan, line 18: `filePath` should probably be named `relPath` and `name` should be named `filePath`.

does `filePath` really need to be resolved?

src/fastly/include-util.js
```diff
@@ -13,6 +13,14 @@
 const fs = require('fs-extra');
 const path = require('path');
 
+function regex(content, filePath) {
```
tripodsan, line 16: what does this function do? jsdoc ?

test/block.test.js
```diff
@@ -16,7 +16,9 @@ const assert = require('assert');
 const path = require('path');
 const fs = require('fs-extra');
 
-const re = new RegExp(fs.readFileSync(path.resolve(__dirname, '../layouts/fastly/helix.vcl')).toString().match(/"(.*)".* # block baddies/)[1].replace(/%25/g, '%'));
+const name = path.resolve('src/rgx/block.rgx');
```
tripodsan, line 19: use `__dirname`. otherwise test is depedent on `process.cwd()`

test/block.test.js
```diff
@@ -16,7 +16,9 @@ const assert = require('assert');
...
+const re = new RegExp(fs.readFileSync(name).toString().split('\n').join(''));
```
tripodsan, line 20: same here: use `readFileSync(name, 'utf-8')`

tripodsan (changes requested):
some minor details....

however, it would probably be better to have the RegExp as code instead of a file.

src/fastly/include-util.js
trieloff, line 19:
```suggestion
    const c = fs.readFileSync(name).toString().split('\n').join('|');
```

That way we can skip the trailing pipe in the regex files.

trieloff:
```suggestion
    const c = fs.readFileSync(name)
      .toString()
      .split('\n')
      .map(l => l.replace(/[ ]+#.*$/, '')) // enables comments at the end of the line
      .join('|');
```

even better: comments that can explain each line

src/rgx/block.rgx
```diff
@@ -0,0 +1,59 @@
...
+^/settings\.php\.bak$|^/admin_tdet\.json$|^/phpinfo\.txt$|^/WEB-INF/jetty-env\.xml$|^/Sites/Samples/Knowledge/Search/ViewCode\.json$|^/bigdump\.php$|^/catalog_admin\.sql$|^/uri$|^/admin_edit\.aspx$|^/r\.php$|
+^/gb_admin\.aspx$|^/\.curlrc$|^/rd\.sql$|^/bea_wls_internal/iiop/ClientSend$|^/up\.php$|^/admin-post\.txt$|^/admin2/login\.asp$|^/adminlogon/$|^/sys_log/$|^/dialog/oauth/$|^/admin-header\.aspx$|
+^/admloginuser\.db$|^/2019\.zip$|^/member/login\.json$|^/admin\.cgi$|^/panel-administracion/index\.aspx$|^/modules/admin/$|^/moderator/login\.sql$|^/addadmin\.sql$|^/inc/tinymce/$|^/backend_dev/$|
+^/SaveForLater\.json$|^/libraries$|^/database\.csv$|^/asp/$|^/admin_paylog\.txt$|^/phpMyAdmin-3\.1\.2/$|^/phpMyAds/$|^/admin_userdet\.conf$|^/killer\.php$|^/WEB-INF/components\.xml$|^/isapi/$|^/download/history\.csv$|
+^/actions_admin\.aspx$|^/SaveForLater\.db$|^/user\.asp$|^/ad_admin\.sql$|^/l0gs\.txt$|^/\.checkignore$|^/ids_log\.aspx$|^/vb\.rar$|^/orasso$|^/members/login\.json$|^/_webalizer/$|^/exchange/logon\.aspx$|^/admin_censoring\.conf$|
+^/adminlist\.php$|^/wp-admin/post.php$|^/en/.*(([0-9][0-9][0-9][0-9])|[a-zA-Z])%.*
```
trieloff, line 59: split at each pipe that is not in a group

```suggestion
^/adminlist\.php$
^/wp-admin/post.php$
^/en/.*(([0-9][0-9][0-9][0-9])
[a-zA-Z])%.*
```

test/fixtures/tinyrgx.rgx
```diff
@@ -0,0 +1,8 @@
...
+^/settings\.php\.bak$|^/admin_tdet\.json$|^/phpinfo\.txt$|^/WEB-INF/jetty-env\.xml$|^/Sites/Samples/Knowledge/Search/ViewCode\.json$|^/bigdump\.php$|^/catalog_admin\.sql$|^/uri$|^/admin_edit\.aspx$|^/r\.php$|
+^/gb_admin\.aspx$|^/\.curlrc$|^/rd\.sql$|^/bea_wls_internal/iiop/ClientSend$|^/up\.php$|^/admin-post\.txt$|^/admin2/login\.asp$|^/adminlogon/$|^/sys_log/$|^/dialog/oauth/$|^/admin-header\.aspx$|
+^/admloginuser\.db$|^/2019\.zip$|^/member/login\.json$|^/admin\.cgi$|^/panel-administracion/index\.aspx$|^/modules/admin/$|^/moderator/login\.sql$|^/addadmin\.sql$|^/inc/tinymce/$|^/backend_dev/$|
+^/SaveForLater\.json$|^/libraries$|^/database\.csv$|^/asp/$|^/admin_paylog\.txt$|^/phpMyAdmin-3\.1\.2/$|^/phpMyAds/$|^/admin_userdet\.conf$|^/killer\.php$|^/WEB-INF/components\.xml$|^/isapi/$|^/download/history\.csv$|
+^/actions_admin\.aspx$|^/SaveForLater\.db$|^/user\.asp$|^/ad_admin\.sql$|^/l0gs\.txt$|^/\.checkignore$|^/ids_log\.aspx$|^/vb\.rar$|^/orasso$|^/members/login\.json$|^/_webalizer/$|^/exchange/logon\.aspx$|^/admin_censoring\.conf$|
+^/adminlist\.php$|^/wp-admin/post.php$
```
trieloff, line 8:
```suggestion
^/adminlist\.php$|^/wp-admin/post.php$   # that's all
```

adding a comment to test the comment removal feature

trieloff (commented):
Having comment support would be nice

src/fastly/include-util.js
MarquiseRosier: I use sync because the rest of the code does! I can change this if you'd like! thanks for the tip about using 'utf-8'! :)

src/rgx/block.rgx
```diff
@@ -0,0 +1,59 @@
+([^A-Za-z0-9-_./+%25]|\/\.\.|%25[^2-7a-fA-F][^0-9a-fA-F]|%25(2[26]|3([aA]|[dD])|5[cC]|7[cC])|//etc/|
```
stefan-guggisberg, line 1: It would significantly improve readability and maintainability if the regex would be split one ORed expression per line and sorted alphabetically.

trieloff:
@MarquiseRosier are you still working on this or can I take over?

src/rgx/block.rgx
```diff
@@ -1,48 +1,30 @@
+.swf$
```
stefan-guggisberg, line 1:
```suggestion
\.swf$
```
?

stefan-guggisberg (approved):

src/rgx/block.rgx
```diff
@@ -12,7 +12,6 @@
 /wp-(app|admin|config|contents?|includes|login|rss.*|register)
 /zimbra
 %25(2[26]|3([aA]|[dD])|5[cC]|7[cC])
-%25[^2-7a-fA-F][^0-9a-fA-F]
```
dominique-pfister: This expression is to block requests with '%' not followed by two valid hexadecimal characters, as this again provokes a 500 and message `Illegal query: Invalid input 'g', expected HEXDIG (line 1, column 313)`.

Can we add an example to the block list for this instead?

dominique-pfister: Noticed that we actually need two expressions:
```
%25[^2-7a-fA-F]             # match % followed by anything but sensible hexadecimal
%25[2-7a-fA-F][^0-9a-fA-F]  # match % followed by sensible hexadecimal but not another hexadecimal
```

tripodsan: I think we can simplify the regexp to only allow lowercase letters, numbers, `-` and `_`.
at least that's what we discussed with @davidnuescheler.

so we don't need this super complex regexp anymore.

stefan-guggisberg: hmm... paths like 

`/foo/bar%1`
`/foo/bar%`

aren't blocked but still trigger `500`s

tripodsan: if we only allow letters, digits and dashes, `/foo/bar%1` would be blocked.

stefan-guggisberg: > I think we can simplify the regexp to only allow lowercase letters, numbers, - and _.
at least that's what we discussed with @davidnuescheler.

Didn't we had cases just recently where the path did contain `%20` (live site)?

tripodsan: :-) user error....

ok, allow lowercase letters, -, _, ., and `%20`. but that's it.

dominique-pfister: > so we don't need this super complex regexp anymore.

which one? 😄

stefan-guggisberg: > I think we can simplify the regexp to only allow lowercase letters, numbers, `-` and `_`

no extension?

dominique-pfister: > I think we can simplify the regexp to only allow lowercase letters, numbers, - and _

If this relates to **all** request URLs for theblog, we already have at least 4 documents that won't work, because their URL contains %-encoded characters.

trieloff: @dominique-pfister please add the examples directly to the branch

dominique-pfister: Done, these were taken from https://github.com/adobe/theblog/issues/260

trieloff:
We can discuss the suggested simplification and it's merits in #592

PR #591, opened 2020-08-27
fix(purge): ignore url params for hlxpurge

### LINKED ISSUE #589: Double check query parameter handling
After talking with @rofe I'd like to double check the handling of URL parameters is correct. Ideal behavior: 
- innercdn purges all urls with all query parameters
- outer purges just the exact URL

I think at the moment just one of the two conditions is met and I don't know which one

trieloff:
fixes #589

## Related Issues

Thanks for contributing!

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -625,7 +625,7 @@ sub hlx_type_purge {
     + "/purge@v1"
     + "?host=" + urlencode(req.http.X-Orig-Host)
     + "&xfh="  + urlencode(req.http.X-Forwarded-Host)
-    + "&path=" + urlencode(regsuball(req.http.X-Orig-Url, "\?.*$", ""));
+    + "&path=" + urlencode(req.http.X-Orig-Url);
 
   set req.request = "POST";
 }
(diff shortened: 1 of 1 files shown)


layouts/fastly/helix.vcl
```diff
@@ -625,7 +625,7 @@ sub hlx_type_purge {
     + "/purge@v1"
     + "?host=" + urlencode(req.http.X-Orig-Host)
     + "&xfh="  + urlencode(req.http.X-Forwarded-Host)
-    + "&path=" + urlencode(req.http.X-Orig-Url);
+    + "&path=" + urlencode(regsuball(req.http.X-Orig-Url, "\?.*$", ""));
```
tripodsan, line 628: wouldn't it be better to pass in the original url, and trim the query string in helix-purge when needed?
the outer CDN still needs the query string to flush by url, no?

trieloff: Good point.

```suggestion
    + "&path=" + urlencode(req.http.X-Orig-Url);
```

stefan-guggisberg (approved):

rofe (approved):

PR #592, opened 2020-08-28
Simple Regular Expression

trieloff:
Only letters, numbers dashes.

See #577


src/rgx/block.rgx
```diff
@@ -0,0 +1 @@
+[^a-z0-9-_/\.]   # only lowercase letters, numbers, dashes as @uncled seggested. Also / and . because we're friendly
```
stefan-guggisberg, line 1:
```suggestion
[^a-z0-9-_/\.]   # only lowercase letters, numbers, dashes as @uncled suggested. Also / and . because we're friendly
```

tripodsan (approved):

stefan-guggisberg (approved):

tripodsan:
@trieloff why did you close this?

trieloff:
- incompatible with #597 
- generates too many false positives for the blog
- does not catch half of the true positives in the examples

overall, an idea worth trying out, but not worth pursuing, given the results.

tripodsan:
> * generates too many false positives for the blog

like what?

trieloff:
https://app.circleci.com/pipelines/github/adobe/helix-publish/2098/workflows/4ed36e95-ad4e-4037-91a1-f1d27f74a22f/jobs/3323

```
  645) Test allowed paths
       allow /en/publish/2020/04/30/AG-Test.html:

      AssertionError [ERR_ASSERTION]: 1 false match for /[^a-z0-9-_\/\.]/
      + expected - actual

      -1
      +0
```

tripodsan:
well, of course we have to educate :-)

maybe we can reconsider this, once https://github.com/adobe/helix-word2md/issues/358 is fixed.

tripodsan:
@davidnuescheler some questions:
- the `_` is not allowed in the 2nd segment... is this intentional ?
- the `.` is not allowed in the leading segment, which blocks `.well-known/acme-challenge`...is this ok?

stefan-guggisberg (approved):

test/fixtures/allowed-paths.txt
```diff
@@ -459,7 +436,6 @@
 /hlx_be96d7f08b8a2c907a09572f8bae21b9b1d0d3a1.jpg
 /lessons/assets/focus-ring.png
 /lessons/welldone.html
-/README.html
```
trieloff, line 462: This makes it really hard to work with existing sites.

test/fixtures/allowed-paths.txt
```diff
@@ -947,7 +915,6 @@
 /hlx_c7ea45c7441987b3c228059e87e2ba445dfd41b7.jpg
 /hlx_e34a64cdabd444792699c4e703495c377639307a.jpg
 /hackathons/3-sfo.html
-/docs/Helix%20Architecture.png
```
trieloff, line 950: Come on, a little space has never hurt anyone.

test/fixtures/allowed-paths.txt
```diff
@@ -1022,7 +988,6 @@
 /hlx_e5dcc1db3f1f11bebf7fc35de9f6f199560076bc.jpeg
 /hlx_26909954f61aa359412f1e8f024678c2850116de.jpeg
 /hlx_ee505e436bf92041978210dc3082fc708e4837f7.jpg
-/office/%E6%97%A5%E6%9C%AC.html
```
trieloff, line 1025: What do you have against Korean?

test/fixtures/blocked-paths.txt
```diff
@@ -186,371 +186,81 @@
 /en/drafts/migrated/202045739960%20or%209285%3d09285--%20/02/26/need-to-save-9-1m-in-three-years-digitize-your-document-processes.html
 /en/drafts/migrated%7cping%20-c%2021%20127.0.0.1%7c%7cx/2019/12/10/december-photography-releases.html
 /~admin/
-/test_ip.php
```
trieloff, line 189: PHP is always malicious (except in `cgi-bin`, where it's cool)

test/fixtures/blocked-paths.txt
```diff
@@ -186,371 +186,81 @@
...
-/test5.php
-/test6.php
-/test7.php
-/test8.php
-/test9.php
-/testadmin
-/testproxy.php
-/tests/phpunit_report.xml
-/testweb
-/text-base/etc/passwd
```
trieloff, line 204: Obvious attack

test/fixtures/blocked-paths.txt
```diff
@@ -186,371 +186,81 @@
...
 /thank-you.1.marketing.adobe.com.txt
 /thanks.1.marketing.adobe.com.txt
 /ThankYou.1.marketing.adobe.com.txt
-/theme
-/themes
 /themes/default/htdocs/flash/ZeroClipboard.swf
-/thirdparty/fckeditor
```
trieloff, line 211: Obvious attack

trieloff (commented):
Hm. I can't "request changes" on my own PR, so a "Comment" will have to do. 

The editing of the test files to match the results of the regular expression seems like an inadequate tactic to cover the inadequacy of this approach.

I do think that an allow list to complement the block list could work, but we likely need both and the allow list must cast a wider net.

test/fixtures/allowed-paths.txt
```diff
@@ -1060,8 +1024,4 @@
...
-/cgi-bin/test.js?id=JJJ0QQQ&action=JJJ1QQQ&page=JJJ2QQQ&name=JJJ3QQQ&f=JJJ4QQQ&url=JJJ5QQQ&email=JJJ6QQQ&type=JJJ7QQQ&file=JJJ8QQQ&title=JJJ9QQQ&code=JJJ10QQQ&q=JJJ11QQQ&user=JJJ12QQQ&token=JJJ13QQQ&t=JJJ14QQQ&c=JJJ15QQQ&data=JJJ16QQQ&mode=JJJ17QQQ&order=JJJ18QQQ&lang=JJJ19QQQ&p=JJJ20QQQ&key=JJJ21QQQ&status=JJJ22QQQ&start=JJJ23QQQ&charset=JJJ24QQQ&s=JJJ25QQQ&post=JJJ26QQQ&login=JJJ27QQQ&search=JJJ28QQQ&content=JJJ29QQQ&comment=JJJ30QQQ&step=JJJ31QQQ&ajax=JJJ32QQQ&debug=JJJ33QQQ&state=JJJ34QQQ&query=JJJ35QQQ&error=JJJ36QQQ&save=JJJ37QQQ&sort=JJJ38QQQ&format=JJJ39QQQ&tab=JJJ40QQQ&offset=JJJ41QQQ&edit=JJJ42QQQ&preview=JJJ43QQQ&filter=JJJ44QQQ&from=JJJ45QQQ&view=JJJ46QQQ&a=JJJ47QQQ&limit=JJJ48QQQ&do=JJJ49QQQ&plugin=JJJ50QQQ&theme=JJJ51QQQ&text=JJJ52QQQ&test=JJJ53QQQ&path=JJJ54QQQ&pass=JJJ55QQQ&dir=JJJ56QQQ&show=JJJ57QQQ&h=JJJ58QQQ&value=JJJ59QQQ&filename=JJJ60QQQ&redirect=JJJ61QQQ&year=JJJ62QQQ&group=JJJ63QQQ&template=JJJ64QQQ&subject=JJJ65QQQ&m=JJJ66QQQ&u=JJJ67QQQ&dest=JJJ68QQQ&uri=JJJ69QQQ&continue=JJJ70QQQ&window=JJJ71QQQ&next=JJJ72QQQ&reference=JJJ73QQQ&site=JJJ74QQQ&&W0SrC=JJJ75QQQ&
```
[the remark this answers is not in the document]
tripodsan: I think the query params should not be considered in the allow/block lists.

test/fixtures/allowed-paths.txt
tripodsan: we don't support unicode.... nobody uses it anyways.

tripodsan:
> I do think that an allow list to complement the block list could work, but we likely need both and the allow list must cast a wider net.

I don't think we need both... I just reversed the logic, because it was easier to incorporate @davidnuescheler's regexp.

tripodsan (changes requested):
let's find consensus first....

test/fixtures/allowed-paths.txt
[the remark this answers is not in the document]
trieloff: We have to. Remember, the main reason we use these filters is not to prevent attacks, but to prevent origin requests that only end up in 404s. Normally, we would strip all URL parameters from pipeline requests and let the cache handle it, but with the helix-pages disabling of caches we can't use that anymore.

There are certain tells in URL parameters like `../../` or `etc/passwd` that make good filters.

trieloff:
I dislike the goalpost-moving done in: https://github.com/adobe/helix-publish/pull/592/commits/953a90a57ad0369a684f874a2877ee46bd449c22 everything in `allowed-paths.txt` that you are removing is a real path that is breaking. Everything in `blocked-paths.txt` that you are removing is a real attack you are letting slip though.

I'm all for @davidnuescheler's idea of having a customizable allowlist of root paths, but I'm having a hard time of seeing the merits of this approach.

tripodsan:
> I dislike the goalpost-moving done in: [953a90a](https://github.com/adobe/helix-publish/commit/953a90a57ad0369a684f874a2877ee46bd449c22) everything in `allowed-paths.txt` that you are removing is a real path that is breaking. Everything in `blocked-paths.txt` that you are removing is a real attack you are letting slip though.
> 
> I'm all for @davidnuescheler's idea of having a customizable allowlist of root paths, but I'm having a hard time of seeing the merits of this approach.

the amount of false positive denies with the current regexp is much higher than the amount of false positive blocks.
i.e. having to update the  regexp every time a new fancy pattern appears is also not very fruitful.

davidnuescheler:
@stefan-guggisberg and i are trialing the variant described by @trieloff on a couple of our production outer CDN config, so far things look promising... we can update this issue once we are done with a significant part of the rollout.

tripodsan:
closing won't fix.

PR #597, opened 2020-08-31
fix(regexp): use JS regexp as source

### LINKED ISSUE #596: using file for regexp doesn't work for in deployed action
**Description**

since the `block.rgx` file is not available as file when the action is webpacked, it doesn't work:

```
error: Remote publish service failed Error: 500 - "{
  "message": "Error: ENOENT: no such file or directory, open '/nodejsAction/m0v0omYq/src/rgx/block.rgx'",
  "stack": ["Error: ENOENT: no such file or directory, open '/nodejsAction/m0v0omYq/src/rgx/block.rgx'", "    at Object.openSync (fs.js:443:3)", "    at Object.readFileSync (fs.js:343:35)", "    at content.replace (/nodejsAction/m0v0omYq/main.js:66371:18)", "    at String.replace (<anonymous>)", "    at regex (/nodejsAction/m0v0omYq/main.js:66369:18)", "    at Object.init (/nodejsAction/m0v0omYq/main.js:66788:13)", "    at publish (/nodejsAction/m0v0omYq/main.js:66993:11)", "    at process._tickCallback (internal/process/next_tick.js:68:7)"],
  "status": "error"
}"
```

## Suggest:

- turn the file into a `.js` file, which returns the final `Regexp` and require it.

tripodsan:
fixes #596


trieloff (changes requested):
Isn't #598 the less invasive approach?

tripodsan:
> Isn't #598 the less invasive approach?

yes...but this is better. as it uses JS regexp with enables proper syntax coloring and allows for normal comments.
I already mentioned this in:  https://github.com/adobe/helix-publish/pull/577#pullrequestreview-469915728 (but it was ignored).

trieloff (approved):
ok

PR #601, opened 2020-09-02
feat(helix.vcl): enable caching of blocked requests

### LINKED ISSUE #600: Return 404 instead of 403 for blocked requests
Requests with url paths matching the "bad request" regex currently return status `403 - Forbidden`. This poses a problem for outer CDNs as status 403 is by default [not cached by Fastly](https://docs.fastly.com/en/guides/http-status-codes-cached-by-default). 

**Describe the solution you'd like**
Such requests should rather return 404 which is by default cached by CDNs.

**Additional context**

stefan-guggisberg:
fix #600

## Related Issues

Thanks for contributing!

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -212,7 +212,7 @@ sub hlx_block_recv {
   set req.http.X-Trace = req.http.X-Trace + "; hlx_block_recv";
 
   if (!req.http.x-topurl && req.url.path ~ {"regex:block.rgx"}) { # block baddies
-    error 955 "Forbidden";
+    error 951 "Not Found";
   }
 
   call hlx_deny;


tripodsan (approved):

trieloff (changes requested):
That will make it hard for us to assess the proportion of malicious requests.

You could create a new synthetic response path that looks like a 404, but adds a `req.http.X-Status = 403;` and then we'd adjust the log schema to use `req.http.X-Status` for `cdn.status` with higher prio than `resp.status`.

stefan-guggisberg:
hmm, sounds overly complicated. 

BTW: there are still quite a lot of baddies which aren't caught by the regex. They already return 404.

davidnuescheler:
after just reviewing the logs the of a real life 24h, i can definitely support that the "malicious requests" are currently returning a mix 403 and 404's alike and i really don't see any value in reporting an arbitrary portion of those as 403, i find it actually quite annoying when looking through the logs that i have to query both to get a complete picture.

the bigger problem is that we have false positives (or negatives depending how you look at it) with the current regex, which surface as a 403 for legitimate URLs, which is where a 404 would be very hard to diagnose for the user. 
however, the main problem is the complexity of the regex, so we should really fix that.

tripodsan:
@stefan-guggisberg @trieloff @davidnuescheler can we close this `won't fix` ?

PR #627, opened 2020-10-05
Preflight Support

### LINKED ISSUE #349: Preflight Conditions
With the new conditions language in `master`, we can take the next step: enable conditions based on a preflight HTTP request.

## Use Case

A certain strain should be made available to visitors with certain properties that cannot be evaluated by the conditions language itself, for instance because it requires some form of database lookup and cookie matching. Typical scenarios are using products like Adobe AudienceManger or Adobe Target to decide which audience a visitor belongs to.

## Configuration

```yaml
conditions:
  and:
    url: https://www.mywebsite.com
    delegate:
      service: https://target.adobe.io/mbox
      x-audience: 25171
```

In the example above, the condition would only be true if the request URL starts with `https://www.mywebsite.com` and a forwarded request to `https://target.adobe.io/mbox` has a header `x-audience` matching `25171`.

## Constraints

Preflight requests are a relative costly way of determining strain membership because:

- they need to be made before any of the other strain resolution happens
- they need to be made for every request that does not have a strain attached to it
- they need to be made as soon as any strain is using a `delegate` condition
- they need to be made for every combination of `delegate/service` in the entire configuration
- they need to be made in sequence (latency)
- each `delegate/service` combination counts towards the (hard) limit of `restart;`statements
- each `delegate/service` combination counts towards the (soft) limit of backends

We should therefore institute a limit of 2 `delegate/service` pairs per Helix service.

## Considerations

- [ ] should the strain resolution be moved from the shield to the edge, as there isn't anything cachable here?
- [ ] can delegates on the edge take advantage of distributed "edge" services for matching

- if evaluation of the body of the delegate service response is required, then a proxy service (on Adobe I/O Runtime or a low-latency edge compute engine) would be necessary.

## Implementation Sketch

As provided by @ejthurgo in https://fiddle.fastlydemo.net/fiddle/67780242:

`recv`:
```vcl
if (req.restarts < 1) { # NOTE: the hard coded number 1 would be increased if multiple services are supported
  log "syslog foo someloggingendpoint :: " + "performing pre-flight" + "" + ;
  # Clear out the X-Strain header to prevent spoofing
  unset req.http.X-Strain;
  
  # Stash the original URL
  set req.http.X-Orig-Url = req.url;
  set req.http.X-Preflight = "1";

  #   Update backend
  set req.backend = F_delegate_1;
  
  # Ideally we would cache the response here, but what do we vary on?
  return(pass);
  
} else {
  log "syslog foo someloggingendpoint :: " + "going to real origin with x-strain: " + req.http.X-Strain + ;
  
  # Reinstate the original URL
  set req.url = req.http.X-Orig-Url;
  
  # Disable pre-flight
  unset req.http.X-Preflight;
  
  # Do something with the strain?
  set req.url = querystring.add(req.url, "strain", req.http.X-Strain);
}

#TODO: Re-enable shielding following updates to backend.
```

`fetch`
```vcl
set beresp.ttl = 1d;
return(deliver);
```

`deliver`:
```vcl
# Interrogate preflight response, add flags to req and restart
if (req.http.X-Preflight) {

  # Extract first header – this would need to be repeated for each unique header in any delegate condition
  if (resp.http.x-audience) {
    set req.http.delegate-1-x-audience = respect.http.x-audience;
  }
  
  restart;
}
```

The condition fragment to be used in strain resolution would then look like this:

```vcl
if (req.http.delegate-1-x-audience ~ "^25171") {
  # set the strain…
}
```

## Related Issues

This is an update and refinement of adobe/helix-publish#5 and replaces it.

trieloff:
Fixes #349 

Depends on #626


trieloff:
Try it:

`http https://www.primordialsoup.life/index.html` vs. `https://www.primordialsoup.life/index.html` return different, but fully cached results based on the user agent.

from: https://github.com/trieloff/helix-demo/blob/f38cb7042522cd0e00a3690d7e9ccfe62afd0857/helix-config.yaml#L1-L15
```yaml
preflight: https://adobeioruntime.net/api/v1/web/trieloff/10e05c072b108ac9533c6206c6eaa651901ca29b/cgi-bin-strain


strains:
  - name: new
    condition:
      preflight.site-version>: 1
    content: 'https://github.com/trieloff/helix-demo.git#new-things'
    static: https://github.com/trieloff/helix-demo.git/htdocs#master
    code: ssh://git@github.com/trieloff/helix-demo.git#master
    package: trieloff/10e05c072b108ac9533c6206c6eaa651901ca29b
  - name: default
    content: 'https://github.com/trieloff/helix-demo.git#master'
    static: https://github.com/trieloff/helix-demo.git/htdocs#master
    code: ssh://git@github.com/trieloff/helix-demo.git#master
    package: trieloff/10e05c072b108ac9533c6206c6eaa651901ca29b
```

The `preflight` script at https://adobeioruntime.net/api/v1/web/trieloff/10e05c072b108ac9533c6206c6eaa651901ca29b/cgi-bin-strain will return different values for the `Site-Version` header based on the `User-Agent` header. (see https://github.com/trieloff/helix-demo/blob/master/cgi-bin/strain.js)

We can then use the value of this header to select a strain, in this case we just check if the version is greater than 1 (`preflight.site-version>: 1`)

---

This PR has a been a long time coming (and I don't just mean the month+ it took from opening it as a draft to getting it ready for review) – the idea (#5) goes back to 2018 (SF hackathon) – and it would not have been possible without a little help from my friends:

- @ejthurgo provided the first draft implementation at the Berkeley hackathon
- @dominique-pfister built our conditions support that I could easily hook into
- @stefan-guggisberg made the `X-Trace` feature so robust that I could understand what was going on
- @tripodsan turned this from a neat idea into a requirement by needing it to manage breaking changes in helix-pages

Issue #662, opened 2020-12-01
Read action variables from version-lock strain property


tripodsan:
**Is your feature request related to a problem? Please describe.**
currently, the following backend action versions are hardcoded:

```
  dispatch: 'v4',
  content-proxy: 'v2',
  purge: 'v1',
  query-index: 'v1',
  static: 'v1',
```

**Describe the solution you'd like**
The publish action should store the versions in the strain dictionary and the VCL should use the versions from the dictionaries.

PR #663, opened 2020-12-01
feat(vlc): use version-lock header and edge dictionary

tripodsan:
see #662


trieloff (changes requested):
I'd try create a single dictionary for the version lock, load it (unless the header has been set) and parse the value on demand using [`subfield`](https://developer.fastly.com/reference/vcl/functions/miscellaneous/subfield/) – building this in the review box is a bit tricky, we can connect tomorrow morning via Slack.

src/fastly/vcl.js
```diff
@@ -81,10 +81,11 @@ function updatestrains(fastly, version, strains, config) {
   ]);
 }
 
-function dynamic(fastly, version, dispatchVersion) {
+function dynamic(fastly, version, defaultVersions) {
   let content = '';
   content += xversion(version, package.version);
-  content += reqHeader('X-Dispatch-Version', dispatchVersion);
+  content += reqHeader('X-Dispatch-Version', defaultVersions.dispatch);
```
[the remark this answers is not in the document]
tripodsan: well, I just kept it backward compatible, because the dispach version can be specified as action parameter for helix-publish.

src/fastly/include-util.js
```diff
@@ -44,7 +45,7 @@ function injectConsts(content, constants) {
     .entries(constants)
     .filter(([key, value]) => !!key && !!value)
     .reduce((str, [key, value]) => str
-      .replace(`{"const:${key}"}`, `"${value}"`),
+      .replace(new RegExp(escapeStringRegexp(`{"const:${key}"}`), 'g'), `"${value}"`),
```
tripodsan, line 48: need to use regexp replace  with global flag, since nodejs doesn't have a `String.replaceAll()`

layouts/fastly/helix.vcl
```diff
@@ -327,6 +327,20 @@ sub hlx_github_static_root {
...
+sub hlx_version_lock {
+  set req.http.X-Trace = req.http.X-Trace + "; hlx_version_lock";
+  if (req.http.X-OW-Version-Lock) {
+    set req.http.X-Trace = req.http.X-Trace + "(header)";
+    return;
+  }
+
+  set req.http.X-OW-Version-Lock =
+    table.lookup(strain_version_lock, req.http.X-Strain, // fallback to the strain-defined version lock
+      table.lookup(strain_version_lock, "default") // fallback to the default strain
```
tripodsan, line 340: there is no default version lock header to be set, so there is no fallback.

tripodsan:
@trieloff I redid the PR:

- there is now only a `strain_version_lock` dictionary, containing the version lock header, computed from the strain config
- all the defaults for the internal actions (dispatch, static, etc) are now constants
- a respective entry in the strain/version_lock can override the default version
- only the strain defined version lock will be set on the `x-ow-version-lock` header (we don't need to send along all other defaults)
- the dispatch version can still be set via the action parameter (don't know if we still need this)

TODO: I still need to test if this all works :-)

trieloff:
> the dispatch version can still be set via the action parameter (don't know if we still need this)

I think we should drop this (BREAKING) as there is a more convenient workaround.

src/fastly/vcl.js
```diff
@@ -48,9 +48,21 @@ async function init(fastly, version, options, config) {
...
+  const constants = {
     preflight: config.preflight ? config.preflight.replace('https://adobeioruntime.net', '') : undefined,
+  };
+
+  // define default version constants
+  const defVersions = {
+    ...defaultVersions,
+    ...versions,
+  };
+  Object.values(defVersions).forEach((k, v) => {
```
trieloff, line 62: I'd use `reduce` over `forEach`

src/fastly/vcl.js
```diff
@@ -48,9 +48,21 @@ async function init(fastly, version, options, config) {
...
+    constants[`${k}_version`] = v;
   });
```
trieloff, line 64:
```suggestion
  // define default version constants
  const defVersions = {
    ...defaultVersions,
    ...versions,
  };
  const constants = Object.entries(defVersions).reduce([c, (k, v)] => {
    c[`${k}_version`] = v;
    return c;
  }, {
    preflight: config.preflight ? config.preflight.replace('https://adobeioruntime.net', '') : undefined,
  });
```

trieloff (approved):

src/fastly/vcl.js
tripodsan: why? it only makes it look more complicated

tripodsan:
> > the dispatch version can still be set via the action parameter (don't know if we still need this)
> 
> I think we should drop this (BREAKING) as there is a more convenient workaround.

let's do this in a separate PR, currently this is controllable in hlx cli:

```
  --dispatch-version, --dispatchVersion     Version of the dispatch action to use.                        [string]
```

layouts/fastly/helix.vcl
```diff
@@ -327,6 +327,20 @@ sub hlx_github_static_root {
...
+  set req.http.X-Trace = req.http.X-Trace + "; hlx_version_lock";
+  if (req.http.X-OW-Version-Lock) {
+    set req.http.X-Trace = req.http.X-Trace + "(header)";
+    return;
+  }
+
+  set req.http.X-OW-Version-Lock = table.lookup(strain_version_lock, req.http.X-Strain);
+  if (req.http.X-OW-Version-Lock == "") {
+    set req.http.X-OW-Version-Lock = table.lookup(strain_version_lock, "default");
+  }
```
tripodsan, line 341: @trieloff I need to use this variant instead of the table default value fallback, because it never worked for me. i think the fallback only works, if the entry is missing, but in our case, the entry has the empty string value.

also, the `if (!req.http.X-OW-Version-Lock)` variant also didn't work.... not sure why.

layouts/fastly/helix.vcl
```diff
@@ -1220,9 +1264,16 @@ sub hlx_type_dispatch {
   # get (strain-specific) parameter allowlist
   include "params.vcl";
 
+  # get dispatch version
+  declare local var.dispatch_version STRING;
+  set var.dispatch_version = subfield(req.http.X-OW-Version-Lock, "dispatch", "&");
+  if (var.dispatch_version == "") {
+    set var.dispatch_version = {"const:dispatch_version"};
+  }
```
tripodsan, line 1272: @trieloff same here: the `if (!var.dispatch_version) {` syntax didn't work....

trieloff (approved):

PR #671, opened 2020-12-07
feat(preflight): include repo/owner/ref

### LINKED ISSUE #670: Preflight requests should include repo/owner/ref
When implementing a version picker, the github repository is required to read additional information.
right now, the owner/repo/ref information is removed from the bereq request.

tripodsan:
fixes #670

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -1213,10 +1213,17 @@ sub hlx_type_content {
 sub hlx_type_preflight {
   set req.http.X-Trace = req.http.X-Trace + "; hlx_type_preflight";
 
+  call hlx_owner;
+  call hlx_repo;
+  call hlx_ref;
+
   # get it from OpenWhisk (for now, we will support other backends later)
   set req.backend = F_AdobeRuntime;
 
-  set req.http.X-Backend-URL = {"const:preflight"};
+  set req.http.X-Backend-URL = {"const:preflight"}
+      + "?owner=" + req.http.X-Owner
+      + "&repo=" + req.http.X-Repo
+      + "&ref=" + req.http.X-Ref;
 }
 
 /**
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -1220,10 +1220,7 @@ sub hlx_type_preflight {
   # get it from OpenWhisk (for now, we will support other backends later)
   set req.backend = F_AdobeRuntime;
 
-  set req.http.X-Backend-URL = {"const:preflight"}
-      + "?owner=" + req.http.X-Owner
-      + "&repo=" + req.http.X-Repo
-      + "&ref=" + req.http.X-Ref;
+  set req.http.X-Backend-URL = {"const:preflight"};
 }
 
 /**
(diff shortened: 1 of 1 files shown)


layouts/fastly/helix.vcl
```diff
@@ -1213,10 +1213,17 @@ sub hlx_type_content {
...
+  call hlx_ref;
+
   # get it from OpenWhisk (for now, we will support other backends later)
   set req.backend = F_AdobeRuntime;
 
-  set req.http.X-Backend-URL = {"const:preflight"};
+  set req.http.X-Backend-URL = {"const:preflight"}
+      + "?owner=" + req.http.X-Owner
+      + "&repo=" + req.http.X-Repo
+      + "&ref=" + req.http.X-Ref;
```
trieloff, line 1226: Don't you think it would be enough to have this in the headers? That way the cache efficiency stays high, as you'd only `Vary` on the headers you actually need?

trieloff (commented):
Adding the headers is good, using url parameters less so.

layouts/fastly/helix.vcl
tripodsan: but it's different for each repository picked repository. if we add owner/repo/ref as headers, we also need to vary on all of them..

I also thought of using the x-helix-forwarded-url header, and compute the coords; but this would then be less robust.

trieloff: The advantage of using headers and vary is that you get exactly the granularity you need, `x-helix-forwarded-url` would be too fine grained.

layouts/fastly/helix.vcl
```diff
@@ -1781,23 +1785,25 @@ sub hlx_bereq {
   if (req.http.X-Request-Type != "Preflight") {
     unset bereq.http.X-Orig-Url;
     unset bereq.http.X-Orig-Host;
+    unset bereq.http.X-Owner;
+    unset bereq.http.X-Repo;
+    unset bereq.http.X-Ref;
+    # unset bereq.http.X-Repo-Root-Path;   // todo: should we remove this as well ?
```
tripodsan, line 1791: @trieloff this was not removed from the bereq before... is it important to the backends ?

layouts/fastly/helix.vcl
```diff
@@ -1781,23 +1785,25 @@ sub hlx_bereq {
...
-  unset bereq.http.X-Repo;
-  unset bereq.http.X-Ref;
   unset bereq.http.X-Root-Path;
   unset bereq.http.X-Action-Root;
   unset bereq.http.X-Github-Static-Repo;
   unset bereq.http.X-Github-Static-Owner;
   unset bereq.http.X-Github-Static-Root;
   unset bereq.http.X-Github-Static-Ref;
   unset bereq.http.X-Restarts;
+  # unset bereq.http.X-Trace; // todo: should we remover this as well ?
```
tripodsan, line 1806: @trieloff this was not removed from the bereq before... is it important to the backends ?

layouts/fastly/helix.vcl
```diff
@@ -1788,23 +1785,25 @@ sub hlx_bereq {
   if (req.http.X-Request-Type != "Preflight") {
     unset bereq.http.X-Orig-Url;
     unset bereq.http.X-Orig-Host;
+    unset bereq.http.X-Owner;
+    unset bereq.http.X-Repo;
+    unset bereq.http.X-Ref;
```
trieloff, line 1790:
```suggestion
```

@tripodsan  Potentially. I'd leave the headers in so we don't break anything.

layouts/fastly/helix.vcl
trieloff:
```suggestion
  unset bereq.http.X-Trace; // todo: should we remover this as well ?
```

Good idea to remove this.

trieloff (approved):

layouts/fastly/helix.vcl
tripodsan: ?? I just moved the `unset`s into the if. they were removed before.

trieloff: Nevermind then.

layouts/fastly/helix.vcl
trieloff: No, but for things like the cgi-bin it could be useful

PR #704, opened 2021-01-26
Universal Runtime Support

trieloff:
Adds support for the universal runtime.

Request types:
- [x] `dispatch`
- [x] `static_url`
- [x] `static`
- [x] `purge`
- [x] `cgi`
- [x] `query`
- [x] `embed`
- [ ] ~~`preflight`~~ will be done later, needs support in helix-shared.
- [x] `content`

@tripodsan for `static_url` `static` `purge` `query` `embed` we rely on the OpenWhisk linking feature that we have not replicated yet in `helix-dispatch`. Should we just hardcode a hostname (a.k.a. namespace) such as `helix-shared.anywhere.run`? We'd need to make sure the corresponding actions are deployed to this service.


tripodsan:
> for `static_url` `static` `purge` `query` `embed` we rely on the OpenWhisk linking feature that we have not replicated yet in `helix-dispatch`. Should we just hardcode a hostname (a.k.a. namespace) such as `helix-shared.anywhere.run`? We'd need to make sure the corresponding actions are deployed to this service.

I don't quite understand the problem. the action invokes stay in the same environment.

eg: a request to:

https://helix-pages.anywhere.run/helix-services/dispatch@v4

will invoke the dispatch on the `helix-pages` namespace, which will then use the `__OW_API_HOST` and `__OW_NAMESPACE` to invoke the subsequent actions.

see: https://github.com/adobe/helix-deploy/blob/7e21dc39969dadbd224d661b43240c948ded28c0/src/template/resolver.js#L94-L95

trieloff:
Let's take `dispatch` and `purge` as examples. 

`dispatch` works just as you described, Fastly requests `https://helix-pages.anywhere.run/helix-services/dispatch@v4` so everything is fine.

Now, `purge` is different:

1. `purge` could be `https://helix-pages.anywhere.run/helix-services/purge@v1`, but this assumes that
  - either that the helix-pages project deployed *all* helix-related services to it's own namespace (on OW) and API gateway (on AWS)
  - or that the `helix-services` package is linked on OpenWhisk and the `helix-pages` API gateway is somehow mounting the lambdas from the `helix` AWS account into the `helix-pages` API gateway

or:
2. `purge` could be `https://helix-services.anywhere.run/helix-services/purge@v1`, but this comes with the drawback of:
  - no neat service sharing and accounting of invocations
  - all `helix-services` invocations count against the same rate limits

tripodsan:
we don't have namespaces in AWS, and all helix-services are mounted in the `helix-pages` namespace in OW.

> 2. purge could be https://helix-services.anywhere.run/helix-services/purge@v1, ...

no. that would be wrong.

trieloff:
We deploy helix-pages (html action, etc) and helix-services (e.g. purge) differently in OW. I'd assume that we would have the same separation in AWS, i.e. different subdomains for anywhere.run. The question is how to reconcile this here.

tripodsan:
> We deploy helix-pages (html action, etc) and helix-services (e.g. purge) differently in OW. I'd assume that we would have the same separation in AWS, i.e. different subdomains for anywhere.run. The question is how to reconcile this here.

ah, ok. in AWS we just ignore the namespaces, and _hope_ there is no collision in the packages :-)
at a later stage, we could choose different API Gateways for different namespaces, and then use different AWS resource groups to execute the functions, for separation of limits/costs (I guess)

trieloff:
Ok, for now I will assume the specified subdomain has all the actions and later we can build https://github.com/adobe/helix-deploy/issues/91

tripodsan (approved):
hmm..there is a lot of duplicated code that calculates the universal/namespace vars. can't this be done in a function?

trieloff:
VCL does not have function, it only has subroutines, which makes this difficult. Ideally, we'd have functions that return a struct with `var.namespace` and `var.universal` that we could simply call. I thought about building something that would inline a subroutine, but decided to go with the more obvious, but also more verbose way.

stefan-guggisberg (approved):
seems ok (but to be  honest: I don't really understand it)

layouts/fastly/helix.vcl
```diff
@@ -545,6 +545,32 @@ sub hlx_type_static_url {
...
+  # them into the URL
+  declare local var.universal BOOL;   # use universal runtime?
+  declare local var.hostname STRING;  # if yes, what's the hostname
+  declare local var.namespace STRING; # namespace
+  declare local var.package STRING;   # package
+
+  # We need the action root for the next bit
+  call hlx_action_root;
+
+  if (req.http.X-Action-Root ~ "(^|^https://)([^/:\.]+)(/|\.([^/]+)/)([^/]+)") {
```
trieloff, line 558: This pattern supports both formats of the `package` parameter: `namespace/package` and `https://namespace.anywhere.run/package`

layouts/fastly/helix.vcl
```diff
@@ -545,6 +545,32 @@ sub hlx_type_static_url {
...
+    set var.universal = false;
+    set var.namespace = re.group.2;
+    set var.package = re.group.5;
+
+    if(re.group.1 == "https://") {
+      set var.universal = true;
+      set var.hostname = var.namespace + "." + re.group.4;
+      set req.backend = F_UniversalRuntime;
+      set req.http.X-Backend-Host = var.hostname;
```
trieloff, line 567: We are using the universal runtime, so we need to store the backend hostname, e.g. `namespace`

layouts/fastly/helix.vcl
```diff
@@ -555,8 +581,8 @@ sub hlx_type_static_url {
     set var.static_version = {"const:static_version"};
   }
 
-  set req.http.X-Action-Root = "/api/v1/web/" + table.lookup(secrets, "OPENWHISK_NAMESPACE") + "/helix-services/static@" + var.static_version;
-  set req.http.X-Backend-URL = req.http.X-Action-Root
+  set req.http.X-Backend-URL = if(var.universal, "/", "/api/v1/web" + "/" + var.namespace)
```
trieloff, line 584: Add the `/api/v1/web/${namespace}` prefix for OpenWhisk – Universal Runtime does not need it.

layouts/fastly/helix.vcl
```diff
@@ -1761,6 +1910,19 @@ sub hlx_bereq {
       } else {
         set bereq.http.hlx-forwarded-host = req.http.X-Orig-Host;
       }
+    } elsif (req.backend == F_UniversalRuntime) {
+      // Set the hostame for the universal Runtime service
+      set bereq.http.Host = req.http.X-Backend-Host;
```
trieloff, line 1915: Use the backend hostname set here https://github.com/adobe/helix-publish/pull/704/files#diff-9e59debf5608a9bdb1f1a87b891b2f995823ea79254a122b8d1fcb2bb1f69785R567 (for instance)

As `*.anywhere.run` will always go to Fastly, setting the subdomain name here allows Fastly to pick the correct gateway service.

trieloff (commented):
@stefan-guggisberg I've added some elaboration.

PR #718, opened 2021-02-03
Support .hlx virtual resources to allow rootless setups

trieloff:
See https://github.com/adobe/fedpub/issues/12 when running Helix Pages at a non-root location behind a CDN, all our `/_hlx/…` resources are broken because they are at an inaccessible path.

This PR will introduce a set of `*.hlx` virtual resources that work at any path depth

- Images from Word and Google will be at `/foo/blob.hlx/79f1216a4ab875f5d4a7f6e2ab77cc337700bf.png`
- CGI-BIN will be at `/foo/cgi-bin.hlx/scriptname.pl`
- Queries will be at `/foo/query.hlx/indexname/queryname`
- Fonts will be at `/foo/fonts.hlx/4odo83.css`

Related:
- adobe/helix-static#485: makes sure Typekit CSS looks correct and uses relative paths
- adobe/helix-pipeline#955: makes sure the pipeline creates relative image links

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -830,6 +830,12 @@ sub hlx_type_query {
     set var.qindex_version = {"const:query-index_version"};
   }
 
+  # Only declare local variables for things we mean to change before putting
+  # them into the URL
+  declare local var.universal BOOL;   # use universal runtime?
+  declare local var.hostname STRING;  # if yes, what's the hostname
+  declare local var.namespace STRING; # namespace
+  
   # We need the action root for the next bit
   call hlx_action_root;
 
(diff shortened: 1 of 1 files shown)


davidnuescheler:
i think that we have a range of URLs that become problematic in cases where helix doesn't completely control the URL namespace of a host.

i would like to use this opportunity (where we are breaking backwards compatibility) to carefully select an intuitively and semantically correct naming convention that makes sense for someone who looks at the URL to understand what to expect.

i think we should test some of the use cases / assertions for the categories below and decide if we want to continue support for those. so let me comment on the individual cases.

> Images from Word and Google will be at /foo/blob.hlx/79f1216a4ab875f5d4a7f6e2ab77cc337700bf.png
to me this is the most interesting and obivously the most broadly used case.
i think this is bigger than images, as we also upload rich media via the slack bot.
so in my mind this category should be called `media` or `asset`
personally, i would love to have them on the same path level as the referencing html.
so as an example, a post of: 


> CGI-BIN will be at /foo/cgi-bin.hlx/scriptname.pl
> Queries will be at /foo/query.hlx/indexname/queryname
> Fonts will be at /foo/fonts.hlx/4odo83.css

trieloff:
I'll try again next week. Thanks for the input.

PR #726, opened 2021-02-09
fix(preflight): do not process preflight requests for external embeds

### LINKED ISSUE #725: preflight requests loose headers in ESI requests
the preflight requests that were don't via an ESI request, cause errors in the version picker and leak the error response into the ESI response:

In the example below, the `owner, repo, ref required.` is a `400` response from the version picker.
the ESI is from the pipeline that injects the embed-service.

image

trieloff:
The `embed` request type is quite different from other request types, as it goes to a different host, so the typical method of determining owner, repo, ref does not work. On the other hand, the underlying service does not make use of strains at all, so there is litte point in trying to resolve a strain. This change will skip preflight requests for embeds. A side effect is that the version-picker for helix-embed won't work for users in helix-pages.

fixes #725

### DIFF
--- a/src/fastly/vcl-utils.js
+++ b/src/fastly/vcl-utils.js
@@ -123,7 +123,7 @@ function resolve(mystrains, preflight) {
 # Strain resolution depends on preflight completion. If the preflight request has
 # not been made, we need to start it now.
 
-if (req.restarts < 1 && !req.http.X-Request-Type) {
+if (req.restarts < 1 && !req.http.X-Request-Type && req.http.host != "adobeioruntime.net") {
   set req.http.X-Request-Type = "Preflight";
 } else {
 `;


tripodsan (commented):
but isn't the embed url the embed-service ?
https://github.com/adobe/helix-pipeline/blob/2ff56e4a5e31ddb5b1896a537c746be9fd273559/src/utils/embed-handler.js#L29

trieloff:
Yes, the embed URL is the embed service (`https://adobeioruntime.net/api/v1/web/helix/helix-services/embed@v1``) which will show up in VCL as `req.http.host == "adobeioruntime.net"` and `req.url == "/api/v1/web/helix/helix-services/embed@v1`"`

All other requests have the actual hostname in `req.hostname`.

tripodsan (approved):

PR #750, opened 2021-02-18
feat(strains): allow strain selection via x-stealthy-strain header

trieloff:
other than the explicit selection via the x-strain header, this one does not affect the cachability. This also means that changes to x-stealty-strain will only become effective once the cache has expired

see https://github.com/adobe/helix-pages/pull/695#issuecomment-781187574

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -168,7 +168,7 @@ sub hlx_strain {
     set req.http.X-Trace = req.http.X-Trace + "(cookie)";
   }
   # read strain from stealty-strain header
-  if (req.http.X-Stealty-Strain) {
+  if (req.http.X-Stealthy-Strain) {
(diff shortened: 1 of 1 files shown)


layouts/fastly/helix.vcl
```diff
@@ -167,6 +167,11 @@ sub hlx_strain {
     set req.http.X-Strain = req.http.Cookie:X-Strain;
     set req.http.X-Trace = req.http.X-Trace + "(cookie)";
   }
+  # read strain from stealty-strain header
+  if (req.http.X-Stealty-Strain) {
```
tripodsan, line 171:
```suggestion
  if (req.http.X-Stealthy-Strain) {
```

although, I think `X-Strain-No-Vary` is more descriptive....

tripodsan (approved):

PR #791, opened 2021-03-12
fix(vcl): handle unknown methods

### LINKED ISSUE #790: Non GET requests do not get rejected
https://github.com/adobe/helix-publish/blob/2950616c7afb4eda8fc43dea1018d0021a483123/layouts/fastly/helix.vcl#L1656-L1659

This code, added by @drwilco two years ago does not seem to work (anymore?) Maybe it is because we don't handle the 405 with a custom response?

trieloff:
fixes #790

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -1119,6 +1119,8 @@ sub hlx_fetch_error {
       error 965 "Too many requests";
     } elsif (beresp.status == 400) {
       error 966 "Bad request";
+    } elsif (beresp.status == 400) {
+      error 967 "Method not allowed";
     } else {
        error 952 "Internal Server Error";
     }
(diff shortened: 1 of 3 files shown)


layouts/fastly/helix.vcl
```diff
@@ -1119,6 +1119,8 @@ sub hlx_fetch_error {
       error 965 "Too many requests";
     } elsif (beresp.status == 400) {
       error 966 "Bad request";
+    } elsif (beresp.status == 400) {
```
stefan-guggisberg, line 1122: duplicate if stmt

stefan-guggisberg (changes requested):

layouts/fastly/helix.vcl
dominique-pfister: shouldn't this be 405?

layouts/fastly/helix.vcl
```diff
@@ -1744,7 +1746,12 @@ sub hlx_deliver_errors {
 
   if (resp.status == 966) {
      set resp.status = 400;
-     set resp.response = "Bad request";
+     set resp.response = "Bad Request";
+  }
+
+  if (resp.status == 967) {
+     set resp.status = 400;
```
dominique-pfister, line 1753: and again 405?

dominique-pfister (approved):

PR #803, opened 2021-03-25
fix(vcl): allow wider blob hash

tripodsan:
for the new blob hashes we need to allow longer sequences.

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -814,7 +814,7 @@ sub hlx_type_blob {
   declare local var.ext STRING;
   declare local var.sas STRING;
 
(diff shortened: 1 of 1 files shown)


trieloff (changes requested):
There is at least one other occurrence, search for `if (req.url.path ~ "^/hlx_(([0-9a-f]){40}).([0-9a-z]+)$" || req.url.path ~ "/media_(([0-9a-f]){40}).([0-9a-z]+)$") {`

kptdobe:
This is a pretty serious change, are we sure about the regex ?

tripodsan:
> This is a pretty serious change, are we sure about the regex ?

I'm sure about the regexp, but I'm not sure if we don't have other locations we look for 40 characters.
but we'll find out quickly. this PR is probably not a problem.

trieloff (approved):

PR #832, opened 2021-04-13
feat(vcl): allow passing the x-github-token header in request

trieloff:
up to this change, the github token would always be taken from a globally shared token, making it impossible to have truly private repositories for websites as the helix user would need to be a member of the repository, effectively making it world readable.

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -1879,6 +1879,11 @@ sub vcl_fetch {
   # Vary on x-ow-version-lock, to avoid caching of different versions
   set beresp.http.Vary:X-OW-Version-Lock = "";
 
+  # Vary on X-Github-Token, to avoid caching secret content
+  if (req.backend == F_AdobeRuntime || req.backend == F_UniversalRuntime || req.backend == F_GitHub) {
+    set beresp.http.Vary:X-Github-Token = "";
+  }
+
 
   if (beresp.http.Expires || beresp.http.Surrogate-Control ~ "max-age" || beresp.http.Cache-Control ~ "(s-maxage|max-age)") {
     # Use TTL from origin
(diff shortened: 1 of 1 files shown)


layouts/fastly/helix.vcl
```diff
@@ -1879,6 +1879,11 @@ sub vcl_fetch {
   # Vary on x-ow-version-lock, to avoid caching of different versions
   set beresp.http.Vary:X-OW-Version-Lock = "";
 
+  # Vary on X-Github-Token, to avoid caching secret content
+  if (req.backend == F_AdobeRuntime || req.backend == F_UniversalRuntime || req.backend == F_GitHub) {
+    set beresp.http.Vary:X-Github-Token = "";
```
tripodsan, line 1884: we don't need this, since the github token would only be used for request to the content proxy.

layouts/fastly/helix.vcl
```diff
@@ -1999,9 +2004,15 @@ sub hlx_bereq {
     # set Adobe Runtime backend authentication
     set bereq.http.Authorization = table.lookup(secrets, "OPENWHISK_AUTH");
     # pass Github Token via X-Github-Token header
-    set bereq.http.X-Github-Token = table.lookup(secrets, "GITHUB_TOKEN");
+    if (!req.http.X-Github-Token) {
+      # use global token if header not set
+      set bereq.http.X-Github-Token = table.lookup(secrets, "GITHUB_TOKEN");
```
tripodsan, line 2009: I think having a github token per repository stored in the table would be better. something like:

```
set bereq.http.X-Github-Token = table.lookup(githubtoken, repo)
```

tripodsan (changes requested):
I disagree. passing a github token into `*.page` should not have any effect.
the pages service should use a distinct table to lookup the tokens.

tripodsan:
also, passing in the token from `*.live` is not practicable. otherwise `*.pages` can be used for preview or local development.

@rofe @davidnuescheler WDYT?

trieloff:
> otherwise `*.page` can't be used for preview or local development. 

The simulator can be extended to pass the GitHub token, that solves local development.

For preview, you can set up a `*.hlx.dev` domain with IP ACLs.

layouts/fastly/helix.vcl
trieloff: But you would be able to make a request to an `.md` file that has been delivered before and get the cached version, without any authentication. In fact, you could poison the cache by requesting the URL with a missing or wrong token and the 404 would be cached.

layouts/fastly/helix.vcl
trieloff: That leaves us with the problem of keeping the table updated, which would not scale well and cause issues with key rotation, taking sites offline, etc. It also shifts the frontier of customization from the outer CDN to the inner CDN.

layouts/fastly/helix.vcl
tripodsan: but this feature is not about authentication in .page, but how to access private content. if .page automatically provides the token for the content, the cache is always correct.

as I said: IMO a x-github-token passed to .page should not have any effect.

layouts/fastly/helix.vcl
tripodsan: currently there is 1 customer.

tripodsan:
> The simulator can be extended to pass the GitHub token, that solves local development.

local dev uses .page. and as I said above, not github token.

> For preview, you can set up a `*.hlx.dev` domain with IP ACLs.

the IP acls should be on .page

tripodsan (approved):

PR #838, opened 2021-04-17
fix(vcl): set vary headers on synthetic (error) responses

### LINKED ISSUE #836: Set Vary on Synthetic Errors

trieloff:
fixes #836

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -1758,6 +1758,17 @@ sub hlx_deliver_errors {
      set resp.status = 400;
      set resp.response = "Bad request";
   }
+
+  set resp.http.Vary:X-Debug = "";
+  set resp.http.Vary:X-Strain = "";
+  set resp.http.Vary:X-Request-Type = "";
+  # why vary unconditionally here? All error pages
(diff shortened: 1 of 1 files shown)


layouts/fastly/helix.vcl
```diff
@@ -1758,6 +1758,17 @@ sub hlx_deliver_errors {
      set resp.status = 400;
      set resp.response = "Bad request";
   }
+
+  set resp.http.Vary:X-Debug = "";
```
tripodsan, line 1762: out of curiosity: setting the header to `""`, does it enable the vary on x-debug or disable it?

tripodsan (approved):

layouts/fastly/helix.vcl
trieloff: The `*.http.Header-Name:Compontent-Name = "Value";` syntax is used to access cookie values (`Key1=value1,key2=value2`) so setting a compontent without any value will just add the key to the string (or keep it there if already set). Conveniently, this is exactly the syntax the `Vary` header needs, so `set ... = ""` enables it, and `unset ...` disables it.

stefan-guggisberg (approved):

Issue #830, opened 2021-04-13
Set CORS headers for known Helix origins


rofe:
In order to allow cross-origin requests between Helix domains (e.g. sending a `HLXPURGE` or `POST` request to `https://foo--bar.hlx.live` from `https://foo--bar.hlx.live` (see also https://github.com/adobe/helix-pages/issues/854), we would need to set the `Access-Control-Allow-Origin` and `Access-Control-Allow-Methods` headers. 

Ideally we would set these on the inner CDN and let them bubble up to the outer CDN(s). 

Pseudo VCL: 
```
if (req.http.Origin` ~ "*.hlx.page" || req.http.Origin ~ "*.hlx.live") {
  set obj.http.Access-Control-Allow-Origin = req.http.Origin;
  set obj.http.Access-Control-Allow-Methods = "OPTIONS, POST, HLXPURGE";
}
```

(The `Access-Control-Allow-Methods` should ideally prevent sending the `Access-Control-Allow-Origin` header in a response that may be cached downstream, so we could avoid having to `Vary` on `Origin`.)

See also https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

trieloff:
Who is setting the `Origin` request header? Is this supposed to be an XFH? I feel like the pseudo VCL is a bit too pseudo to make things clear.

tripodsan:
> Who is setting the `Origin` request header? Is this supposed to be an XFH? I feel like the pseudo VCL is a bit too pseudo to make things clear.

the browser

trieloff:
We'd still need a good way to manage the allowlist of domains.

rofe:
> We'd still need a good way to manage the allowlist of domains.

Why? I think whenever it ends with `hlx.page` or `hlx.live` we can safely add the complete origin.

rofe:
A working POC for this solution is already implemented on the acom-reviews service, and a semi-working one on pages.adobe.com.

rofe:
Which probably means we should only set the headers if not already present to allow for customization in the outer CDN.

trieloff:
Helix Publish does not hardcode any domain it is running on and I'd like to keep it that way.

rofe:
Why do you think it matters which domain it is running on?

trieloff:
The `if (req.http.Origin ~ "*.hlx.page" || req.http.Origin ~ "*.hlx.live") {` in your example gives the impression.

rofe:
[...]

PR #869, opened 2021-05-17
feat(content): add origin vary and cors headers to .lnk requests

trieloff:
fixes (maybe) #830

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -1106,7 +1106,7 @@ sub hlx_fetch_lnk {
     if (!req.is_esi_subreq) {
       set beresp.http.X-PostFetch = beresp.http.X-PostFetch + "; hlx_fetch_lnk";
     }
-    if (req.url.qs ~ "hlx_report=true") {
+    if (req.http.X-Orig-URL ~ "hlx_report=true") {
       set beresp.http.Access-Control-Allow-Origin = req.http.Origin;
       set beresp.http.Access-Control-Allow-Methods = "GET, OPTIONS, POST";
       set beresp.http.Vary:Origin = "";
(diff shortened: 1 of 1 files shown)


layouts/fastly/helix.vcl
```diff
@@ -1101,6 +1101,19 @@ sub hlx_fetch_query {
...
 }
 
+sub hlx_fetch_lnk {
+  if (req.http.X-Request-Type == "Content/LNK") {
+    if (!req.is_esi_subreq) {
+      set beresp.http.X-PostFetch = beresp.http.X-PostFetch + "; hlx_fetch_lnk";
+    }
+    if (req.http.X-Orig-URL ~ "hlx_report=true") {
+      set beresp.http.Access-Control-Allow-Origin = req.http.Origin;
+      set beresp.http.Access-Control-Allow-Methods = "GET, OPTIONS, POST";
```
rofe, line 1111:
```suggestion
      set beresp.http.Access-Control-Allow-Methods = "GET, OPTIONS";
```
`POST` is not needed for .lnk requests

rofe (approved):

layouts/fastly/helix.vcl
```diff
@@ -2359,11 +2374,14 @@ sub vcl_deliver {
     set req.http.x-last-activation-id = resp.http.x-last-activation-id;
   }
 
-  if (!req.http.X-Debug) {
-    # Unless we are debugging, shut up chatty headers
+  if (!req.http.X-Debug && req.http.X-Request-Type != "Content/LNK") {
```
tripodsan, line 2377: why would be ever unset those? 
right now, the content proxy correctly sets the cors headers. so no need to recompute them here.

layouts/fastly/helix.vcl
tripodsan: I would remove all this and just allow the CORS headers from the backend to be honoured.

tripodsan (changes requested):

layouts/fastly/helix.vcl
trieloff: We are serving from a number of backends that are all opinionated about CORS, including Adobe I/O and GitHub. The generic cleanup is the best approach to deal with that.

trieloff:
Obsolete, check #871 instead.

PR #871, opened 2021-05-18
fix(content): allow cors headers for lnk request type

### LINKED ISSUE #830: Set CORS headers for known Helix origins

trieloff:
Alternative to #869 as suggested by @tripodsan – @rofe please test. Fixes #830

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -2359,11 +2359,14 @@ sub vcl_deliver {
     set req.http.x-last-activation-id = resp.http.x-last-activation-id;
   }
 
-  if (!req.http.X-Debug) {
-    # Unless we are debugging, shut up chatty headers
+  if (!req.http.X-Debug && req.http.X-Request-Type != "Content/LNK") {
     unset resp.http.Access-Control-Allow-Headers;
     unset resp.http.Access-Control-Allow-Methods;
     unset resp.http.Access-Control-Allow-Origin;
+  }
+
+  if (!req.http.X-Debug) {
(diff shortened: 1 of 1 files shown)


rofe:
> @rofe please test

Works 👍

rofe (approved):

trieloff:
Ok, so now we just need to get it through the integration test from hell.

rofe:
I'd also appreciate a quick look from @tripodsan.

tripodsan (approved):

PR #894, opened 2021-06-25
fix(vcl): fix 401 to 500 propagation, propagate as 403 instead

### LINKED ISSUE #893: JSON: 401 becomes 500
See https://cq-dev.slack.com/archives/CGG9B79BK/p1624550913282100

- A new repo, `fstab.yaml` exists, but integration user has no access to the sharepoint drive
- request `/`
- requests `metadata.json` behind the scenes
- access to sharepoint fails with a 401 (as expected)
- gets propagated as a 500 (unexpected)

trieloff:
a 401 from the backend was unhandled so far (500), now we propagate it as a 403 (Forbidden) instead of 401 (Unauthorized) as a 401 would ask the browser for credentials (HTTP Basic Auth) that we do not want to handle anyway.

fixes #893

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -1270,6 +1270,10 @@ sub hlx_fetch_error {
       error 965 "Too many requests";
     } elsif (beresp.status == 400) {
       error 966 "Bad request";
+    } elsif (beresp.status == 401) {
+      // we handle a 401 from the backend just like a 403, as there won't be any
+      // chance that the client can re-authenticate.
+      error 955 "Forbidden";
     } else {
        error 952 "Internal Server Error";
     }


dominique-pfister (approved):

stefan-guggisberg:
> as a 401 would ask the browser for credentials

Not for me, I get a plain 401 response in the browser, without being prompted for credentials

layouts/fastly/helix.vcl
```diff
@@ -1270,6 +1270,10 @@ sub hlx_fetch_error {
       error 965 "Too many requests";
     } elsif (beresp.status == 400) {
       error 966 "Bad request";
+    } elsif (beresp.status == 401) {
+      // we handle a 401 from the backend just like a 403, as there won't be any
```
stefan-guggisberg, line 1274: I am not sure about the 401 -> 403.

trieloff: https://datatracker.ietf.org/doc/html/rfc7235#section-3.1

>   The 401 (Unauthorized) status code indicates that the request has not
>   been applied because it lacks valid authentication credentials for
>   the target resource.  The server generating a 401 response **MUST** send
>   a `WWW-Authenticate` header field (Section 4.1) containing at least one
>   challenge applicable to the target resource.

We do not have a challenge that would work in this case. It might be different for private repos and a missing `x-github-token` header, but for the scenario that triggered the bug (Sharepoint folder not shared with Helix) there is nothing the client can do.

PR #906, opened 2021-08-27
fix(embeds): fix broken server-side embeds

trieloff:
two key issues fixed: 1. the `Embed` request type wasn't detected when the hostname pointed to universal runtime 2. the X-Backend-URL wasn't set, so that the wrong function (HTML pipeline) would be invoked and fail

fixes adobe/helix-pages#971

### DIFF
--- a/layouts/fastly/helix.vcl
+++ b/layouts/fastly/helix.vcl
@@ -611,7 +611,7 @@ sub hlx_determine_request_type {
     return;
   }
 
-  if (req.http.host == "adobeioruntime.net" || req.http.host == ".*\.anywhere\.run") {
+  if (req.http.host == "adobeioruntime.net" || req.http.host == ".*.anywhere.run") {
     if (!req.is_esi_subreq) {
       set req.http.X-Trace = req.http.X-Trace + "(embed)";
     }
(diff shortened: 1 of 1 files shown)


kptdobe:
The ran on hlx-2, not on hlx-1. But it has been overwritten in the meantime by something else by some other branch tests. I just re-ran the test to get it back.
Regarding the failing diff tests, it looks like the image hashes are different. Not sure why. @tripodsan ?

kptdobe:
Here is the test page: https://master--theblog--adobe.hlx-3.page/en/publish/2020/05/27/the-connection-by-adobe-advertising-cloud.html (youtube videos here!!!)
And the diff tests: https://app.circleci.com/pipelines/github/adobe/helix-pages/5427/workflows/462ecceb-a7aa-4563-be7c-a7dd0b5d8f3e

tripodsan:
> Regarding the failing diff tests, it looks like the image hashes are different. Not sure why. @tripodsan ?

they fail since a month due to changed hashes. I have no clue why. whenever I download the pages manually, they are identical.

tripodsan (approved):

